PR #3, opened 2022-04-22
feat(js-array-methods): add filter exercise

pixelass:
Migrated from https://codepen.io/neuefische/pen/oNoxJvw

### DIFF
--- a/sessions/js-array-methods/filter/index.js
+++ b/sessions/js-array-methods/filter/index.js
@@ -1,0 +1,40 @@
+console.clear();
+
+const cards = [
+	{
+		id: '1',
+		question: 'How often can I use <header>?',
+		answer: 'As often as you like.',
+		tags: ['html', 'elements', 'basic'],
+	},
+	{
+		id: '2',
+		isBookmarked: false,
+		question: 'How often can I use <aside>?',
+		answer: 'As often as you like.',
+		tags: ['html', 'elements', 'basic'],
+	},
+	{
+		id: '3',
+		isBookmarked: true,
+		question: 'On which types can I use destructuring assignment?',
+		answer: 'On Objects and Arrays',
+		tags: ['js', 'next'],
+	},
+];
+
+// Filter: returns an array of max. same length
+// but with unaltered content
+// is called with a function that returns a 'truthy' or 'falsy' value
+const notBookmarkedCards = cards.filter(card => !card.isBookmarked);
+console.log(notBookmarkedCards);
+
+// 1: Filter the array for cards with the 'html' tag
+// HINT: use the array method "includes"
+const onlyCardsWithTagHTML = null;
+
+// 2: Filter the array for cards with 3 tags.
+const allCardsWith3Tags = null;
+
+// 3: Filter the array for cards with the 'html' and 'js' tag that are also bookmarked.
+const allCardsWithTagsHTMLAndJSThatAreBookmarked = null;
--- a/sessions/js-array-methods/filter/index.js
+++ b/sessions/js-array-methods/filter/index.js
@@ -36,5 +36,5 @@ const onlyCardsWithTagHTML = null;
 // 2: Filter the array for cards with 3 tags.
 const allCardsWith3Tags = null;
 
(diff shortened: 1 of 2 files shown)


sessions/js-array-methods/filter/index.js
```diff
@@ -0,0 +1,40 @@
...
+
+// 1: Filter the array for cards with the 'html' tag
+// HINT: use the array method "includes"
+const onlyCardsWithTagHTML = null;
+
+// 2: Filter the array for cards with 3 tags.
+const allCardsWith3Tags = null;
+
+// 3: Filter the array for cards with the 'html' and 'js' tag that are also bookmarked.
+const allCardsWithTagsHTMLAndJSThatAreBookmarked = null;
```
doemser, line 40: Here is my solution:
https://codesandbox.io/s/adoring-tree-roeog4?file=/index.js

I was not sure about what task 3. is asking for. Is it what you expected?

pixelass: @doemser your solution doesn't make sense:

```js
(card.tags.includes("html") || card.tags.includes("html")) 
```

I did not create the original exercise. @F-Kirchhoff do you remember who created this exercise.

I am also unsure what the question is asking for.

Here's how I understand it:

Create an array of all cards that have "html" AND "js" that are bookmarked.

Example: 

```js
const allCardsWithTagsHTMLAndJSThatAreBookmarked = cards.filter((card) => {
  return (
    (card.tags.includes("html") && card.tags.includes("js")) && card.isBookmarked
  );
});
```

This will return an empty array since none of the cards have "html" AND "js" 🤷

sessions/js-array-methods/filter/index.js
```diff
@@ -0,0 +1,40 @@
...
+		answer: 'On Objects and Arrays',
+		tags: ['js', 'next'],
+	},
+];
+
+// Filter: returns an array of max. same length
+// but with unaltered content
+// is called with a function that returns a 'truthy' or 'falsy' value
+const notBookmarkedCards = cards.filter(card => !card.isBookmarked);
+console.log(notBookmarkedCards);
```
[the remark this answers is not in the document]
pixelass: I agree that "Filter: returns an array of max. same length …" is irritating. 

I did not write this exercise:

Here's my suggestion:


* if the callback returns `true`, we keep the item
* if the callback return `false`, we remove the item
> Note: The callback accepts "truthy" and "falsy" values.  
> We recommend ensuring booleans though.

sessions/js-array-methods/filter/index.js
doemser: > @doemser your solution doesn't make sense: :trollface:
> 
> ```js
> (card.tags.includes("html") || card.tags.includes("html")) 
> ```
> 

Oh, hahaha :D What happened there?! This is what I meant

```js
(card.tags.includes("html") || card.tags.includes("js")) &&
    card.isBookmarked
```

F-Kirchhoff: Here Gregor is right, the solution is an empty array, therefore lets change the exercise as follows:
```suggestion
// 1: Filter the array for cards with the 'html' tag
// HINT: use the array method "includes"
const onlyCardsWithTagHTML = null;

// 2: Filter the array for cards with 3 tags.
const allCardsWith3Tags = null;

// 3: Filter the array for cards with the 'html' or 'js' tag that are also bookmarked.
const allCardsWithTagsHTMLOrJSThatAreBookmarked = null;
```

F-Kirchhoff (changes requested):
LGTM, just change the the last exercise.

doemser:
> LGTM, just change the the last exercise.

I would also suggest changing lines 26-28 as well.

F-Kirchhoff:
> > LGTM, just change the the last exercise.
> 
> I would also suggest changing lines 26-28 as well.

They were indeed unnecessary, I deleted them.

F-Kirchhoff (approved):

macfrei (approved):
Looks good, can be merged, I think! 🚢

PR #10, opened 2022-05-01
feat(react-useeffect-with-fetch): add exercise 01

pixelass:
migrated from https://codesandbox.io/s/exercise-1-useeffect-with-fetch-bmrmnd

PR #11, opened 2022-05-01
feat(react-useeffect-with-fetch): add exercise 02

pixelass:
migraterd from https://codesandbox.io/s/exercise-2-useeffect-with-fetch-9bivur

### DIFF
--- a/sessions/react-useeffect-with-fetch-exercise-02/.codesandbox/workspace.json
+++ b/sessions/react-useeffect-with-fetch-exercise-02/.codesandbox/workspace.json
@@ -1,0 +1,8 @@
+{
+  "responsive-preview": {
+    "Mobile": [320, 675],
+    "Tablet": [1024, 765],
+    "Desktop": [1400, 800],
+    "Desktop  HD": [1920, 1080]
+  }
+}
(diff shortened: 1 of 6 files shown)


PReichetanz (commented):
This is exactly the same exercise as #10 , is this correct or a mistake?

PReichetanz:
> > This is exactly the same exercise as #10 , is this correct or a mistake?
> > To be sure: [Exercise 1](https://github.com/neuefische/web-pool-2022/blob/main/sessions/react-useeffect-with-fetch/exercises.md#exercise-1) is not using a Codesandbox, so we should use the PR of exercise 2 (this very one here).
> 
> I copied them from here: https://github.com/neuefische/web-pool-2022/blob/main/sessions/react-useeffect-with-fetch/exercises.md
> 
> Looks like this is the "Bonus Exercise"
> 
> Considering the naming: I also copied this from the existing exercises. Not having an "Exercise 1" but instead starting with `2` might be confusing.

I see that it might be confusing not to have an exercise 1 within this web-exercise repository, but since our web-pool-sessions repository is our single source of truth, we should stick to the enumeration there.
Especially if there will be more exercises in the future, it would be more confusing if we start here with number one, but, in our web-pool, exercise one does not need a codesandbox.

F-Kirchhoff:
I would suggtest to close this PR and add the exercise text to the other fetch exercise.

PR #12, opened 2022-05-01
feat(react-usestate-and-events-advanced): add exercise 01

pixelass:
migrated from https://codesandbox.io/s/lifting-up-state-in-form-starter-fh2ydq


sessions/react-usestate-and-events-advanced-exercise-01/src/App.js
```diff
@@ -0,0 +1,63 @@
...
+5. Lift up both states from your Form to your App component.
+6. In your App.js, write a function createUser()(name, email)
+where you call the set functions of both states to update them.
+7. Pass createUser() as an `onCreateUser` prop to your Form component.
+Make sure to accept `onCreateUser` in your Form as well.
+8. Inside the Form's handleSubmit(), replace setName() and setEmail()
+with your newly accepted prop `onCreateUser()`.
+Make sure to pass the correct values to this function.
+ */
+
```
PReichetanz, line 31:
```suggestion
import './styles.css';
```

PReichetanz (changes requested):
Only a missing import as far as I can see.

sessions/react-usestate-and-events-advanced-exercise-01/src/App.js
pixelass: I added that to the `index.js` file. I noticed that you import `styles.css` in every file, which is not required and irritating.

https://github.com/neuefische/web-exercises/pull/12/files#diff-8167e26b54664a0fd930fd36dececaeeb592eb1452934ca0dc0c7dd2a2fe5342R4

I "fixed" it in this case but kept the original in the other cases since I noticed the repeating pattern.

This is something that should be fixed for the other sandboxes too.

PReichetanz: I see the point and agree, thank you for this hint!

pixelass: I'll add ist to app , looks like this was a mistake from my side. (probably a mixup, since I migrated many exercises with the same changes for React 18)
The duplicated import does not apply here.

pixelass: @PReichetanz please check the changes and resolve this conversation.

PReichetanz: I checked it, thanks. I will have this conversation in mind regarding the other codesandboxes!

PReichetanz (approved):
There already is a solution for this exercise, [it can be found here](https://codesandbox.io/s/lifting-up-state-in-form-solution-xe3i2k?file=/src/App.js).

I have to admit that I wrote this exercise, so maybe my approval does not count?

sessions/react-usestate-and-events-advanced-exercise-01/src/App.js
```diff
@@ -0,0 +1,64 @@
...
+<p>Name: {name}</p>
+<p>Email: {email}</p>
+```
+
+In your App.js, `name`and `email` are not defined,
+because their states now only live inside the Form component.
+
+4. Comment out `{name}` and `{email}` for the moment.
+5. Lift up both states from your Form to your App component.
+6. In your App.js, write a function createUser()(name, email)
```
F-Kirchhoff, line 23:
```suggestion
6. In your App.js, write a function createUser(name, email)
```

PR #14, opened 2022-05-01
feat(react-usestate-and-events): add exercise 01

pixelass:
migrated from
https://codesandbox.io/s/usestate-onclick-hlfm9w

### DIFF
--- a/sessions/react-usestate-and-events-exercise-01/src/App.js
+++ b/sessions/react-usestate-and-events-exercise-01/src/App.js
@@ -1,0 +1,21 @@
+/*
+Given are two buttons, the first is going to add
+and second one to substract 1.
+1. Write two functions, add() and subtract().
+Log something to the console.
+2. Add your add() and subtract() function
+to the relevant button.
+*/
+
+import './styles.css';
+
+export default function App() {
+	return (
+		<div className="App">
+			<h1>useState and Events</h1>
+			<h2>onClick Exercise</h2>
+			<button>+ 1</button>
+			<button>- 1</button>
+		</div>
+	);
+}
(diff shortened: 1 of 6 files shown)


PReichetanz (dismissed):

PReichetanz (approved):
Note: I wrote the exercise and the solutions.

I still think they work as expected. 

The tasks should be done in two steps (task 1 + 2, after speaking about useState task 3), this is why a provide two solutions: You can find [task 1 + 2 here](https://codesandbox.io/s/usestate-exercise-01-solution-task-1-2-oiz4ci?file=/src/App.js), and [task 3 here](https://codesandbox.io/s/usestate-exercise-01-solution-task-3-8bz96x?file=/src/App.js).

sessions/react-usestate-and-events-exercise-01/src/App.js
```diff
@@ -0,0 +1,21 @@
+/*
+Given are two buttons, the first is going to add
+and second one to substract 1.
+1. Write two functions, add() and subtract().
+Log something to the console.
+2. Add your add() and subtract() function
+to the relevant button.
+*/
+
```
F-Kirchhoff, line 9: I don't understand the task. Should there be a useState that keeps track over the value? Or shoud these functions only log a small text? If its the latter, we should make this more explicit in the text.

PReichetanz: At this point, we did not speak about useState, so the answer is no. I thus think that our students don't need this explicit information since they don't know anything else.

The following third task of this exercise makes it clear, I think:
"As soon as you spoke about useState, implement a count state and save the changes done by your add() and subtract() function to it. Make sure that the count rendered in your App to be visible in the browser."

I still don't know what to do about this PR or whether we should wait to Dominik to update the useState sessions.

PR #35, opened 2022-06-15
feat(js-fetch-and-asyncawait): add exercises

anitaschiller:
I converted both exercises to sandboxes, please review. In order to be able to commit the exercises, I had to deactivate some linting rules, e.g. that object properties have to be written in camelCase.


sessions/js-fetch-and-asyncawait/star-wars-cards/index.html
```diff
@@ -0,0 +1,13 @@
...
+	<head>
+		<meta charset="utf-8" />
+		<meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no" />
+		<title>HTML & CSS & JS</title>
+		<link rel="stylesheet" href="./css/styles.css" />
+	</head>
+	<body>
+		<main id="root"></main>
+	</body>
+	<script src="./js/index.js"></script>
```
macfrei, line 12:
```suggestion
		<link rel="stylesheet" href="./css/styles.css" />
		<script src="./js/index.js" defer></script>
	</head>
	<body>
		<main id="root"></main>
	</body>
```

sessions/js-fetch-and-asyncawait/star-wars-cards/js/index.js
```diff
@@ -0,0 +1,72 @@
...
+
+HINT: You can use the functions `createCharacterCard` and `renderElement` to create those cards, see above for an example!
+*/
+
+// --v-- your code here --v--
+
+function fetchDataAndRender() {
+	fetch() // ?
+		.then() // ?
+		.then(); // ?
```
macfrei, line 54: I would probably add a `catch` here as well.
```suggestion
	fetch() // ?
		.then() // ?
		.then() // ?
		.catch(); // ?
```

macfrei (changes requested):
I really don't understand the color-clue assignments: https://codesandbox.io/s/color-clue-solution-22olbv?file=/js/index.js:665-675
I could not figure out what the goal of this exercise is and because of that I didn't know whether I did it right or wrong. Also, I changed a lot of the CSS and HTML as there were some unusual styling in place (e.g. styling main instead of body).

The star wars card app is a nice exercise to get used to fetch: https://codesandbox.io/s/star-wars-card-solution-sg775i?file=/js/index.js

ThomasKirst:
> I agree, Lene, this exercise requires some time to figure out what the goal is and how it is supposed to work. It contains some 'noisy' computations that might raise questions that distract students from the real task – which is to implement a `fetch` call.

> I really don't understand the color-clue assignments: https://codesandbox.io/s/color-clue-solution-22olbv?file=/js/index.js:665-675
> I could not figure out what the goal of this exercise is and because of that I didn't know whether I did it right or wrong.

PR #36, opened 2022-06-15
feat(js-forms): add form submit exercise

anitaschiller:
The Codepen exercise has been converted to a Sandbox. Please review :)

### DIFF
--- a/sessions/js-forms/form-submit/.eslintrc.json
+++ b/sessions/js-forms/form-submit/.eslintrc.json
@@ -1,0 +1,27 @@
+{
+  "env": {
+    "browser": true,
+    "es2021": true,
+    "node": true
+  },
+  "extends": ["plugin:react/recommended", "xo"],
+  "parser": "@typescript-eslint/parser",
+  "parserOptions": {
+    "ecmaFeatures": {
+      "jsx": true
+    },
+    "ecmaVersion": "latest",
+    "sourceType": "module"
+  },
+  "plugins": ["react", "@typescript-eslint", "prettier"],
+  "rules": {
+    "no-unused-vars": 0,
+    "react/prop-types": 0,
+    "react/react-in-jsx-scope": 0
+  },
+  "settings": {
+    "react": {
+      "version": "detect"
+    }
+  }
+}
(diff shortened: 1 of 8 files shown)


macfrei (commented):
Solution: https://codesandbox.io/s/form-submit-solution-gwko6r?file=/js/index.js

I added a few minor comments and created a solution. I used the `FormData` constructor which we will use in the future according to the recent head coaches meetings.

ThomasKirst:
Thank you very much for the offered solution.

From my point of view, the second version with retrieving the properties and values of the form through `event.target.elements` and their specific assignments to a new object `formData` is easier for the students to understand.

```js
const {name, age, married, hobbies} = event.target.elements;
const formData = {
    name: name.value,
    age: age.value,
    married: married.value,
    hobbies: hobbies.value,
};
```

The first version, based on `FormData`, introduces a bit more complexity and the need for abstraction.
Questions that need to be answered: "What is this `new` keyword, what is a `constructor`?
We put a DOM element (the form) into this newly created object and then try to get the keys and values with `Object.fromEntries(formData)`.

```js
const formData = new FormData(event.target);
const formattedData = Object.fromEntries(formData);
```

In my opinion, `FormData` should be used when attachments are added dynamically or more fields need to be added or removed programmatically, then the object methods can be useful.
For a simple form with four values that we don´t even send anywhere in encoded form, I would consider it unnecessary additional complexity.

A nice addition to the `simple version` could be a quick check on the status of the checkbox (`true` or `false` instead of its always `on` value).
```js
const formData = {
    // …
    married: married.checked,
};
```

aclaes:
When we discussed this topic in the head coaches meeting, we were talking about React and a use case for more experienced developers. I guess we didn't talk (much) about the introduction to JS form handling.

I agree! The "FormData" solution does add some complexity to understand (or simply to accept).

However, there is a central benefit: When the form fields change (add a field, remove a field, rename a field) there is no additional manual change to the JS code required. This simplifies the handling of complex forms, if one accepts the additional complexity of the way how to write it down in JS.

When introducing form handling in JS to students, I would always (!) start with addressing `event.target.elements`. There are major benefits in explaining and understanding forms.

- `.elements` gathers all input fields of the form
- the individual input fields can be addressed via the definition in the `name` attribute
- the `.value` property can be read to get data from a field

Depending on the classes general understanding, I'd show and discuss the "FormData" method afterwards or later in a react session.

ThomasKirst (approved):

PR #39, opened 2022-06-15
feat(react-custom-hooks): add first draft

anitaschiller:
A new exercise for react-custom-hooks which I prepared for one of my sessions. Please review!


macfrei (commented):
Solution: https://codesandbox.io/s/custom-hooks-copy-to-clipboard-solution-n3j1vn?file=/src/App.js

The exercise is easy to understand and enjoyable, however, I did rewrite the app with not an input but a fetch to an API with random cat facts which can be copied. I thought that might be a nicer exercise? I also included styled-components. 

I rewrote it because the use case of copying a password and the html were not really clear to me 😬 . Let's discuss!

PR #46, opened 2022-07-26
Feat/js variables and numbers

aclaes:


sessions/js-variables-and-numbers/social-media-post/css/styles.css
```diff
@@ -0,0 +1,30 @@
...
+	font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Helvetica, Arial, sans-serif,
+		'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol';
+	display: flex;
+	justify-content: center;
+	align-items: center;
+	height: 100vh;
+	background-color: #333;
+}
+
+header {
```
CJAkkuec, line 18: Regarding best practices: do we want to use type selectors, or do we want to use class selectors exclusively (when/if possible)?

sessions/js-variables-and-numbers/social-media-post/js/index.js
```diff
@@ -0,0 +1,22 @@
+console.clear();
+
+/*
+1. Create the data for a little social media post. Create variables for the following data using the const keyword:
```
CJAkkuec, line 4: I'd probably suggest "small" instead of "little" :)

Perhaps you'd also like to clarify that the intent is to create a variable for each data point.
Like so: "Create a const variable for each data point listed below."

Or something.

sessions/js-variables-and-numbers/social-media-post/js/index.js
```diff
@@ -0,0 +1,22 @@
...
+	- the user who created the post
+	- a boolean called isReported
+*/
+
+// --v-- write your code here --v--
+
+// --^-- write your code here --^--
+
+/*
+2. Log all variables to the console. Then increase the likes by one and log the amount of likes again. Adapt your code from 1.1 if necessary.
```
CJAkkuec, line 17: Gramatically speaking, you wouldn't start a sentence with "then". You could just replace the full-stop with a comma.

sessions/js-variables-and-numbers/social-media-post/index.html
```diff
@@ -0,0 +1,16 @@
...
+	<head>
+		<meta charset="utf-8" />
+		<meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no" />
+		<title>Social Media Post</title>
+		<link rel="stylesheet" href="./css/styles.css" />
+		<script src="./js/index.js" defer></script>
+	</head>
+	<body>
+		<header>
+			<h1>Welcome to this Exercise!</h1>
```
CJAkkuec, line 12: Exercise would be in lowercase here. The capitalizing of nouns in English is a little bit ... weird. It wouldn't apply here, though.

sessions/js-variables-and-numbers/simple-calculator/js/index.js
```diff
@@ -0,0 +1,74 @@
...
+*/
+const addButton = document.querySelector('[data-js="add"]');
+const subtractButton = document.querySelector('[data-js="subtract"]');
+const multiplyButton = document.querySelector('[data-js="multiply"]');
+const divideButton = document.querySelector('[data-js="divide"]');
+const exponentButton = document.querySelector('[data-js="exponent"]');
+const moduloButton = document.querySelector('[data-js="modulo"]');
+
+addButton.addEventListener('click', () => {
+	// 1. Add the two operands and store the result in a variable. Log variable's value to the console.
```
CJAkkuec, line 17: In each instance, it'd be "log the variable's[...]" :)

CJAkkuec (approved):
Looking good (as per usual)! I only have some minor language-specific gripes. :)

sessions/js-variables-and-numbers/social-media-post/css/styles.css
aclaes: This exercise is only concerned about the JS and not the HTML or CSS. 

The styling is only used for a note saying the students should open the JS console. There is no further code in HTML or CSS. I guess it's okay for this use case.

There are a couple of such exercises using this note.

sessions/js-variables-and-numbers/simple-calculator/js/index.js
aclaes: Good catch!

sessions/js-variables-and-numbers/social-media-post/index.html
aclaes: Yes! This is somethig we should fix in a couple of exercises using this note

sessions/js-variables-and-numbers/social-media-post/js/index.js
aclaes: Good catch. I just removed the "Then"

sessions/js-variables-and-numbers/social-media-post/js/index.js
aclaes: Good idea

PR #57, opened 2022-08-15
feat(js-forms): add demo

aclaes:
Issue im neuen Format: neuefische/web-curriculum-new-format#19

### DIFF
--- a/sessions/js-forms/demo-start/css/styles.css
+++ b/sessions/js-forms/demo-start/css/styles.css
@@ -1,0 +1,62 @@
+*,
+*::before,
+*::after {
+	box-sizing: border-box;
+	margin: 0;
+	padding: 0;
+}
+
+body {
+	font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Helvetica, Arial, sans-serif,
+		'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol';
+	background-color: lightgray;
+}
+
+main {
+	display: flex;
+	flex-direction: column;
+	gap: 10px;
+	margin: auto;
+	padding: 20px;
+	width: 90%;
+	height: 100vh;
+	background-color: white;
+}
+
+h1 {
+	padding: 5px;
+	background-color: black;
+	color: white;
+}
+
+form {
+	display: flex;
+	flex-direction: column;
+	gap: 10px;
+}
+
+fieldset {
+	display: flex;
+	flex-direction: column;
+	gap: 10px;
+	padding: 10px;
+}
+
+legend {
+	background-color: black;
+	color: white;
+}
+
+button {
+	margin: auto;
+	width: fit-content;
+}
+
+input[type='submit'] {
+	margin: auto;
+	width: fit-content;
+}
+
+textarea {
+	width: fit-content;
+}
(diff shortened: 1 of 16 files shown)


sessions/js-forms/demo-start/css/styles.css
```diff
@@ -0,0 +1,62 @@
...
+}
+
+input[type='submit'] {
+	margin: auto;
+	width: fit-content;
+}
+
+textarea {
+	width: fit-content;
+}
```
macfrei, line 62: I would suggest the following css instead. The layout is a bit easier with grid and the textarea needs `width: 100%;` so that it doesn't bleed out of the form.
```suggestion
*,
*::before,
*::after {
  box-sizing: border-box;
  margin: 0;
  padding: 0;
}

body {
  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Helvetica,
    Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol';
  background-color: lightgray;
}

main {
  display: grid;
  gap: 10px;
  margin: auto;
  padding: 20px;
  width: 90%;
  background-color: white;
}

h1 {
  padding: 5px;
  background-color: black;
  color: white;
}

form {
  display: grid;
  gap: 10px;
}

fieldset {
  display: grid;
  gap: 10px;
  padding: 10px;
}

legend {
  background-color: black;
  color: white;
}

button {
  margin: auto;
  width: fit-content;
}

input[type='submit'] {
  margin: auto;
  width: fit-content;
}

textarea {
  width: 100%;
}
```

macfrei (approved):
Hello :) looks good, I just refactored the CSS a bit.

sessions/js-forms/demo-start/index.html
```diff
@@ -0,0 +1,77 @@
...
+				ever.
+			</p>
+			<form data-js="form">
+				<fieldset>
+					<legend>Personal Information</legend>
+					<label for="first-name">First Name</label>
+					<input
+						type="text"
+						name="firstName"
+						id="first-name"
```
ThomasKirst, line 25: I can already foresee a possible question from a student: "Why do we mix camel/kebap-case, when do we use what and why"?
Could we name the form field´s attributes `name` and `id` the same?

aclaes: Yes - absolutely! If I remember @CJAkkuec correctly, this was meant as an explicit hook to stumble upon in explanation of the form.

The students should see and understand, that it's not advisable to use `name="first-name"`. Therefore this little example was added to the demo.

I think we could change this in the demo. I have no strong opinion about this. 

It's easier to understand for students, but skips the discussion of this topic.

ThomasKirst: Just to clarify: why would you say "it's not advisable to use `name="first-name"`" – because of the possible receiving side / the server, which would prefer the field name in camelCase notation?

aclaes: I my opinion it's not advisable to use kebab-case because students need to understrand the dot notation doesn't work to access the value. 

```js
form.addEventListener('submit', event => {
  event.preventDefault();

  // syntax error :-/
  console.log(
    event.target.elements.first-name.value
  );
});
```

Using camelCase for  the `name` - Attribute rules out this issue.

PReichetanz: I changed `id` and `for` attributes to camelCase as you suggested in this discussion, see the commit b7f1f49307140314b62c828ebab8f4491edd2765.

djfarly (approved):

PR #62, opened 2022-08-19
Feat/css-selectors-demo

PReichetanz:
Add start and end version of the session demo for css-selectors

See PR in neuefische/web-curriculum-new-format#105


ThomasF85 (approved):
Looks good!

PR #67, opened 2022-08-23
feat(exercise): a11y form exercise

PReichetanz:
The focus of this exercise is a11y in a form.

Students need to find and fix accessibility issues in the HTML file.

Any comment is appreciated as well as suggestions what errors we could add ;)

You can find the [start as a Codesandbox here](https://codesandbox.io/s/github/neuefische/web-exercises/tree/feat/html-forms-a11y-exercise/sessions/html-forms/Finding-a11y-errors?file=/README.md).

List of issues:
- `h1` has `class` instead of `id`
- `form` needs `aria-labelledby` instead of `for`
- `fieldset`: `aria-describedby` is missing
- `textarea`: `label` is missing
- `input`s: `id`s do not match with their `labels`

This exercise corresponds with neuefische/web-curriculum-new-format#14.

### DIFF
--- a/sessions/html-forms/Finding-a11y-errors/.eslintrc.json
+++ b/sessions/html-forms/Finding-a11y-errors/.eslintrc.json
@@ -1,0 +1,29 @@
+{
+  "env": {
+    "browser": true,
+    "es2021": true,
+    "node": true
+  },
+  "extends": ["plugin:react/recommended"],
+  "parser": "@typescript-eslint/parser",
+  "parserOptions": {
+    "ecmaFeatures": {
+      "jsx": true
+    },
+    "ecmaVersion": "latest",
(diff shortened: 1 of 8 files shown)


doemser (approved):
Because I unknowingly read the solutions already in the PR, I didn't attach a solution this time, but I do have a few comments.

Spotting the for/id attributes that do not match is reeeaaallly hard, even if you know what to look for.. it is i vs. I :D  Maybe there could also be another combination that doesn't seem to fit together more obvious.

I found the color input, which asks me what color I'm blind to, really funny, but I can imagine that a color blind person might be offended by it.

I've read that placeholders are basically not a11y friendly because they are either not displayed with enough contrast or are incorrectly perceived as input that has already been made .

Would approve this task.

PR #71, opened 2022-08-26
fix/css-flexbox: wrong and missing css classes in profile challenge

F-Kirchhoff:
the css and html was buggy such that the element were not styled properly.


doemser (approved):
looks good to me, lets get this merged

PR #73, opened 2022-08-30
feat(recap-project-2): character count demo

PReichetanz:
One task of the Recap project 2 is to implement a counter of left characters while using an input field.

This codesandbox is meant as a demo we can provide to the students. 

## Links
- [Codesandbox](https://codesandbox.io/s/github/neuefische/web-exercises/tree/feature/recap-project-2-demo-counter/sessions/recap-project-2/character-count-example)
- [Code](https://github.com/neuefische/web-exercises/tree/feature/recap-project-2-demo-counter/sessions/recap-project-2/character-count-example)
- [Corresponding Task in the Draft PR #114](https://github.com/neuefische/web-curriculum-new-format/blob/recap-project-2/sessions/recap-project-2/recap-project-2.md#4-form-field-text-counter)

## Questions
- I tried to use an `onChange` event, but was not able to get it running. Do we get a problem using the `input` event somehow?
- Do we need a readme here as well?

## Belongs to
- neuefische/web-curriculum-new-format#26
- neuefische/web-curriculum-new-format#114

### DIFF
--- a/sessions/recap-project-2/character-count-example/css/styles.css
+++ b/sessions/recap-project-2/character-count-example/css/styles.css
@@ -1,32 +1,28 @@
 *,
 *::before,
 *::after {
-  box-sizing: border-box;
+	box-sizing: border-box;
 }
 
 body {
-  margin: 0;
-  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Helvetica,
-    Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol';
+	margin: 0;
+	font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Helvetica, Arial, sans-serif,
+		'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol';
 }
 
 .form {
-  margin: 20px auto;
-  max-width: 320px;
-  display: grid;
-  justify-items: start;
-  gap: 12px;
+	margin: 20px auto;
+	max-width: 320px;
+	display: grid;
+	justify-items: start;
+	gap: 12px;
 }
 
 .form__label {
-  display: block;
+	display: block;
 }
 
 .form__character-count {
-  font-size: inherit;
-  color: #666;
-}
-
-.hidden {
-  display: none;
+	font-size: inherit;
+	color: #666;
 }
--- a/sessions/recap-project-2/character-count-example/css/styles.css
+++ b/sessions/recap-project-2/character-count-example/css/styles.css
@@ -1,0 +1,32 @@
+*,
+*::before,
+*::after {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Helvetica,
+    Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol';
+}
+
+.form {
+  margin: 20px auto;
+  max-width: 320px;
+  display: grid;
+  justify-items: start;
+  gap: 12px;
+}
+
+.form__label {
+  display: block;
+}
+
+.form__character-count {
+  font-size: inherit;
+  color: #666;
+}
+
+.hidden {
+  display: none;
+}
--- a/sessions/recap-project-2/character-count-example/index.html
+++ b/sessions/recap-project-2/character-count-example/index.html
@@ -1,0 +1,31 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1, shrink-to-fit=no"
+    />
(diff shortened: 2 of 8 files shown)


sessions/recap-project-2/character-count-example/index.html
```diff
@@ -0,0 +1,31 @@
...
+      <label class="form__label" for="question">Question:</label>
+      <textarea
+        maxlength="200"
+        data-js="question"
+        name="question"
+        id="question"
+        cols="24"
+        rows="3"
+      ></textarea>
+      <small data-js="characterCount" class="form__character-count"
```
aclaes, line 24: You don't need the attribute `data-js="characterCount"` here. It's not used in the JS code

sessions/recap-project-2/character-count-example/index.html
```diff
@@ -0,0 +1,31 @@
...
+        >(<span data-js="amountLeft"></span> characters left)</small
+      >
+      <button>Create card</button>
+    </form>
+  </body>
+  <script src="./js/index.js"></script>
```
aclaes, line 30: It was decided, that we like to follow the best practice to include the `<script>` in the HTML head with a `defer` attribute

```html
<head>
    <!-- ... -->
    <script src="./js/index.js" defer></script>
</header>
```

This should probably fixed in the template.

sessions/recap-project-2/character-count-example/js/index.js
```diff
@@ -0,0 +1,15 @@
+console.clear();
+
+const questionElement = document.querySelector('[data-js="question"]');
+const amountLeft = document.querySelector('[data-js="amountLeft"]');
+const maxLength = questionElement.getAttribute('maxLength');
```
aclaes, line 5: Nice idea to read the attribute!

sessions/recap-project-2/character-count-example/js/index.js
```diff
@@ -0,0 +1,15 @@
...
+
+const updateAmountLeft = value => {
+	amountLeft.innerText = value;
+};
+
+updateAmountLeft(maxLength);
+
+questionElement.addEventListener('input', () => {
```
aclaes, line 13: Yes! It's right to use the `input` event here. In JS the `change` event is only triggered, when the form field uses focus, while `input` is triggered with each input.

This differs from react's `onChange`, which behaves like the `input` event in vanilla.

https://reactjs.org/docs/dom-elements.html#onchange

sessions/recap-project-2/character-count-example/index.html
```diff
@@ -0,0 +1,31 @@
...
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1, shrink-to-fit=no"
+    />
+    <title>Character Count Example</title>
+    <link rel="stylesheet" href="./css/styles.css" />
+  </head>
+  <body>
+    <form class="form">
+      <h1>Fix me</h1>
```
aclaes, line 14: We should add a better headerline ;-)

sessions/recap-project-2/character-count-example/index.html
```diff
@@ -0,0 +1,31 @@
...
+      <label class="form__label" for="question">Question:</label>
```
aclaes, line 15: Using "Question" as form field is tiedly coupled to the quiz app. It makes this code more or less the exakt solution for the recap project assignment.

I think we can use something other so that students need to transfer this solution to the quiz-app.

For example: we can label the field  "Personal Message" (and change the `id` and `name` props accordingly)

sessions/recap-project-2/character-count-example/index.html
```diff
@@ -0,0 +1,31 @@
...
+        data-js="question"
+        name="question"
+        id="question"
+        cols="24"
+        rows="3"
+      ></textarea>
+      <small data-js="characterCount" class="form__character-count"
+        >(<span data-js="amountLeft"></span> characters left)</small
+      >
+      <button>Create card</button>
```
aclaes, line 27: I suggest something generice like "Submit" ass button label

sessions/recap-project-2/character-count-example/css/styles.css
```diff
@@ -0,0 +1,32 @@
...
+.form__label {
+  display: block;
+}
+
+.form__character-count {
+  font-size: inherit;
+  color: #666;
+}
+
+.hidden {
```
aclaes, line 30: Do we need the `hidden` class?

sessions/recap-project-2/character-count-example/js/index.js
aclaes, line 5: The attribute is named `maxlength` with lowercase `l`

```suggestion
const maxLength = questionElement.getAttribute('maxlength');
```

aclaes (changes requested):
Nice work! I thnik we don't need a readme file here, since it isn't an exercise, but a short demo

sessions/recap-project-2/character-count-example/css/styles.css
PReichetanz: Not any more, thanks!

aclaes (approved):
Nice 🥳

PR #74, opened 2022-08-30
feat(challenge): added new challenges for css responsive session

CJAkkuec:
These are the challenges for the CSS Responsive session, based on the agreed upon draft.

Also the demo for CSS responsive.

**Example Static**
Code: https://github.com/neuefische/web-exercises/tree/feature/css-responsive/sessions/css-responsive/example-static
CSB: https://codesandbox.io/s/github/neuefische/web-exercises/tree/feature/css-responsive/sessions/css-responsive/example-static

**Example Responsive**
Code: https://github.com/neuefische/web-exercises/tree/feature/css-responsive/sessions/css-responsive/example-responsive
CSB: https://codesandbox.io/s/github/neuefische/web-exercises/tree/feature/css-responsive/sessions/css-responsive/example-responsive

**Demo Start**
Code: https://github.com/neuefische/web-exercises/tree/feature/css-responsive/sessions/css-responsive/demo-start
CSB: https://codesandbox.io/s/github/neuefische/web-exercises/tree/feature/css-responsive/sessions/css-responsive/demo-start

**Demo End**
Code: https://github.com/neuefische/web-exercises/tree/feature/css-responsive/sessions/css-responsive/demo-end
CSB: https://codesandbox.io/s/github/neuefische/web-exercises/tree/feature/css-responsive/sessions/css-responsive/demo-end

**Responsive Website**
Code: https://github.com/neuefische/web-exercises/tree/feature/css-responsive/sessions/css-responsive/responsive-website
CSB: https://codesandbox.io/s/github/neuefische/web-exercises/tree/feature/css-responsive/sessions/css-responsive/responsive-website

**Playground**
Code: https://github.com/neuefische/web-exercises/tree/feature/css-responsive/sessions/css-responsive/playground
CSB: https://codesandbox.io/s/github/neuefische/web-exercises/tree/feature/css-responsive/sessions/css-responsive/playground

## Issues
- neuefische/web-curriculum-new-format#23


PReichetanz:
## Responsive Website Challenge
- [My Solution](https://codesandbox.io/s/challenge-css-responsive-responsive-website-hvvdsk?file=/css/styles.css)
- styles.css, line 127: " Note: understanding what main-content and main-section..." - Do you mean `card-section` here?
- I did not find the part where I was supposed to add my code (styles.css, line 234 sqq.) until I was done (and had a problem with specificity because of the order of my code). Maybe we should add a hint that we want the students to write their code at this specific place?

CJAkkuec:
> ## Responsive Website Challenge
> * [My Solution](https://codesandbox.io/s/challenge-css-responsive-responsive-website-hvvdsk?file=/css/styles.css)
> * styles.css, line 127: " Note: understanding what main-content and main-section..." - Do you mean `card-section` here?
> * I did not find the part where I was supposed to add my code (styles.css, line 234 sqq.) until I was done (and had a problem with specificity because of the order of my code). Maybe we should add a hint that we want the students to write their code at this specific place?

I enjoy that you switched to using grid. 😂  Fair enough though - if it works, it works. 

The thing about media queries is that you'd usually do one per selector and put them right under the selector you want to include in the media query.

Like so

```
.someSelector {
width: 100%;
}

@media (min-width: 600px) {
.someSelector {
width: 50%;
}
}
```

I wasn't sure how to handle prompting the students to write their media queries, and because media queries overwrite your code at specific points, you can just do one media query per breakpoint and just put them at the end, which is what I figured would make sense for that exercise.

So I'd probably suggest we include a hint like "Once you're done investigating the code a little, scroll to the bottom of the file ... etc". What do you think?

CJAkkuec:
> * I was only slightly confused by the first part: `flex-shrink` was not what I expected at first when thinking about responsiveness, but I guess this is exactly the point why we should keep it as part of the challenge :)

Oh, `flex-shrink` wasn't really the intention here, but I understand how the wording is a little confusing. It's really just about percentage as a responsive units. Should I rephrase it?

PReichetanz:
> > * I was only slightly confused by the first part: `flex-shrink` was not what I expected at first when thinking about responsiveness, but I guess this is exactly the point why we should keep it as part of the challenge :)
> 
> Oh, `flex-shrink` wasn't really the intention here, but I understand how the wording is a little confusing. It's really just about percentage as a responsive unit. Should I rephrase it?

I don't think that rephrasing is necessary here, since the students have just heard about how to handle this kind of thing (i.e. percentage as a responsive unit); I did not have the session beforehand, so this was came to my mind.

PReichetanz:
## Demos and Examples
I don't have a comment for them, but I guess the question whether they are good or not rather depend on using them in a session with students and evaluate afterwards.

CJAkkuec:
> ## Demos and Examples
> I don't have a comment for them, but I guess the question whether they are good or not rather depend on using them in a session with students and evaluate afterwards.

We're currently working on a different demo that's based on the playground exercise. So we'd have the puppy website to showcase a before/after and the playground-like demo to visualize what each responsive feature does, etc.

djfarly (approved):

PR #79, opened 2022-09-01
feat(js-forms-2): add demo

aclaes:


PReichetanz (approved):
I really like this example!

The focus can be set to the topic of our session, but students will immediately see a real-world use case.

I did not find any issues 🚀

PR #86, opened 2022-09-06
Feat/js-array-methods/repetition-challenge

PReichetanz:
Challenge for neuefische/web-curriculum-new-format#130

rendered: https://github.com/neuefische/web-exercises/tree/feat/js-array-methods/map-filter-challenge/sessions/js-array-methods/repetition

## Link to Codesandbox
- [Codesandbox](https://codesandbox.io/s/github/neuefische/web-exercises/tree/feat/js-array-methods/map-filter-challenge/sessions/js-array-methods/repetition?file=/README.md)
## Changes
Update challenge:
- fix typos
- adjust to session content (delete `includes` exercise)
- add tests

Fix:
- equal names in some package.json files


CJAkkuec (commented):
Here's my [solution](https://codesandbox.io/s/twilight-water-vmdlmn?file=/index.js)!

One minor note, as per our earlier discussion:
Perhaps we should include a mention of the `join()` method in the handout.

`const allTitlesInOneString = recipes.map((recipe) => recipe.title).join(", ")`
It's not immediately clear what needs to be done to separate the titles as requested.

PReichetanz:
> Here's my [solution](https://codesandbox.io/s/twilight-water-vmdlmn?file=/index.js)! LGTM!
> 
> One minor note, as per our earlier discussion: Perhaps we should include a mention of the `join()` method in the handout.
> 
> `const allTitlesInOneString = recipes.map((recipe) => recipe.title).join(", ")` It's not immediately clear what needs to be done to separate the titles as requested.

I see your point, thanks again!
I would argue not to add `.join()` to the handout, because it is used only in an extra challenge.
This is why I updated the hint how to use `.join()` in this case and add [this link to mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join#joining_an_array_four_different_ways) for further explanation.

What do you think, @CJAkkuec?

CJAkkuec (approved):

PR #93, opened 2022-09-09
JS Structure SpaceRocket Challenge

djfarly:
CSB: https://codesandbox.io/s/github/neuefische/web-exercises/tree/feature/js-structure/sessions/js-structure/spacerocket?file=/js/launchSequence.js

🚀


PReichetanz (approved):
Very nice challenge!

I've just found one typo regarding the explanation of `deploy.js` and its function `deployPayload()`, but that's it.

[Please find my solution here.](https://codesandbox.io/s/space-rocket-solution-01-6k2tvq?file=/js/launchSequence.js)

PR #96, opened 2022-09-14
feat(array-methods-2): animal string challenge

PReichetanz:
Challenge for JS Array Methods 2 Session.

rendered: https://github.com/neuefische/web-exercises/tree/feat/js-array-methods-2/animal-strings-challenge/sessions/js-array-methods-2/animal-strings

- [CSB](https://codesandbox.io/s/github/neuefische/web-exercises/tree/feat/js-array-methods-2/animal-strings-challenge/sessions/js-array-methods-2/animal-strings?file=/README.md)

refers to neuefische/web-curriculum-new-format#131


sessions/js-array-methods-2/animal-strings/README.md
```diff
@@ -0,0 +1,32 @@
...
+- `Array.findIndex`()`
+- `Array.sort`()`
+- `Array.some`()`
+- `Array.every`()`
+- `Array.reduce`()`
+
+Additionally, you will need:
+
+- `String.startsWith`()`
+- `String.endsWith`()`
```
djfarly, line 24: quick formatting fix
```suggestion
- `Array.includes()`
- `Array.find()`
- `Array.findIndex()`
- `Array.sort()`
- `Array.some()`
- `Array.every()`
- `Array.reduce()`

Additionally, you will need:

- `String.startsWith()`
- `String.endsWith()`
```

sessions/js-array-methods-2/animal-strings/index.js
```diff
@@ -0,0 +1,63 @@
...
+// Hint: sort() mutates the original array, which is bad.
+// -> Use animals.slice().sort(...) to make a copy.
+const animalsSortedAlphabetically = null;
+
+// Hint: Guess what? There is a string method called `endsWith()`.
+const anyAnimalEndsWithLetterZ = null;
+
+const everyAnimalHasMoreThanTwoLetters = null;
+
+const valuesForLengthOfAllAnimalNames = [3, 3, 8, 6, 4, 5, 5, 7, 5, 7, 10, 5, 5, 8, 6, 5, 5, 5];
```
MerleMak, line 49:
```suggestion
// Hint: To find out the length of each animal in the array by using the ".length()" function
```

sessions/js-array-methods-2/animal-strings/index.js
```diff
@@ -0,0 +1,63 @@
...
+
+// Hint: Use the variable 'valuesForLengthOfAllAnimalNames' to sum up all letters.
+const sumOfAllAnimalLetters = null;
```
MerleMak, line 52: // The Solution with the suggested change could look like this:
```js
const sumOfAllAnimalLetters = animalStrings.reduce(
  (previousValue, currentValue) => {
    return previousValue + currentValue.length;
  },
  0
);
```
Do you think the array "valuesForLengthOfAllAnimalNames" is necessary for better understanding? If so nvm :)

MerleMak (commented):
We left only one small remark regarding the reduce function. Fun task!! :)

MerleMak:
Great fun, here is our [solution](https://codesandbox.io/s/animal-strings-challenge-solution-hfg5l4)

CJAkkuec:
I really enjoyed this exercise! Here's my [solution](https://codesandbox.io/s/aged-hill-9riiuy?file=/index.js) (including some notes).

Here's an idea how to handle the sum of all string lengths with map:
```
const mappedSum = animalStrings
  .map((animal) => animal.length)
  .reduce((a, b) => a + b);
```

Perhaps it would be worth mentioning that you're looking for ["characters"](https://jkorpela.fi/chars/spaces.html) rather than "letters", seeing as how whitespace is included in the length of a string.

sessions/js-array-methods-2/animal-strings/index.js
PReichetanz: I completely see your point that we should get rid of the `valuesForLengthOfAllAnimalNames` variable; I have implemented it to make the very first usage of `reduce` easier for the students.

I still think that your solution is quite heavy to understand (although it would be a proper one!).

This is why I wrote hints for two (rather easy) options, using the `.length` property either with or without `map` in the first place.

sessions/js-array-methods-2/animal-strings/index.js
PReichetanz: I deleted this variable because of reasons mentioned below.

CJAkkuec (approved):
LGTM 🚀

PR #99, opened 2022-09-15
Feature/css animations demo

F-Kirchhoff:


sessions/css-animations/demo-end/css/styles.css
```diff
@@ -17,98 +17,96 @@ body {
...
+	box-shadow: rgba(0, 0, 0, 0.3) 0px 19px 38px;
+	transition: 0.05s linear;
 }
 
-.button--shadow:hover {
-  box-shadow: rgba(0, 0, 0, 0.3) 0px 19px 38px,
-    rgba(0, 0, 0, 0.22) 0px 15px 12px;
+.button--scale:active {
+	transform: scale(0.9);
+	box-shadow: rgba(0, 0, 0, 0.16) 0px 3px 6px, rgba(0, 0, 0, 0.23) 0px 3px 6px;
```
PReichetanz, line 77: There is an issue (neuefische/web-curriculum-new-format#140) to refactor all CSS styles; it includes removing rgba color styling and replace it with modern rules.
What do you think, could you update it here as well?

PReichetanz (commented):
Looks good to me (even if it is almost only indentation).

One thing I found is the `rgba` color style; @djfarly had the idea to discuss whether we should replace it with more modern variants, but it has yet to be discussed.

Do you think we could use it here already?

sessions/css-animations/demo-end/css/styles.css
djfarly: we'll fix this for all challenges later

PR #100, opened 2022-09-16
feat(array-methods-2): news flash challenge

PReichetanz:
rendered: https://github.com/neuefische/web-exercises/tree/feat/js-array-methods-2/news-flash/sessions/js-array-methods-2/news-flash

[CSB](https://codesandbox.io/s/github/neuefische/web-exercises/tree/feat/js-array-methods-2/news-flash/sessions/js-array-methods-2/news-flash?file=/README.md)

refers to neuefische/web-curriculum-new-format#131


CJAkkuec:
Nice work! Here's my [solution](https://codesandbox.io/s/amazing-khorana-5v179w?file=/index.js). :)

I think it might be a little easier to tell if the second part of the exercise worked if the lengths of the bodies were a little more dramatically different ( - or add a check/test of some sort, something similar to my own maybe). I'm also unsure if we need mock data that's 100 objects heavy at first, then 51 objects after filtering. Perhaps it would be better to have a smaller number after the filtering, like 20 or so?

Just some thoughts. :)

CJAkkuec (approved):

PR #109, opened 2022-09-23
JS-Fetch Session: Demo Start + End

CJAkkuec:
- [Demo Start CSB](https://codesandbox.io/s/github/neuefische/web-exercises/tree/feature/js-fetch/sessions/js-fetch/demo-start)
- [Demo End CSB](https://codesandbox.io/s/github/neuefische/web-exercises/tree/feature/js-fetch/sessions/js-fetch/demo-end)

📌 Rendered: https://github.com/neuefische/web-exercises/tree/feature/js-fetch/sessions/js-fetch

📌 Refers to https://github.com/neuefische/web-curriculum-new-format/issues/36 / neuefische/web-curriculum-new-format#146

### DIFF
--- a/sessions/js-fetch/demo-end/js/index.js
+++ b/sessions/js-fetch/demo-end/js/index.js
@@ -25,7 +25,7 @@ async function getJoke() {
 			// Success
 			const jokeData = await response.json();
 
-			const randomIndex = getRandomInt(0, jokeData.length);
+			const randomIndex = getRandomInt(0, jokeData.length - 1);
 
 			const randomJoke = jokeData[randomIndex];
 
--- a/sessions/js-fetch/demo-end/js/index.js
+++ b/sessions/js-fetch/demo-end/js/index.js
@@ -1,0 +1,40 @@
+// Import Functions & Modules
+import {JokeSection} from '../components/JokeSection/JokeSection.js';
+import {Joke} from '../components/Joke/Joke.js';
+import {getRandomInt} from '../utils/getRandomInt.js';
+
+// Declare root Element
+const root = document.body;
+
+// Assemble DOM
+const jokeSection = JokeSection();
+const joke = Joke();
+
+root.append(jokeSection);
+jokeSection.append(joke);
+
+// Get a random joke from a Bad Jokes API
+async function getJoke() {
+	try {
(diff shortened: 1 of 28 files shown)


PReichetanz:
Thank you @CJAkkuec, this is a really nice demo for our session! 🚀

I am only wondering about the beginning of the `index.js`:
```js
// Declare root Element
const root = document.body;

// Assemble DOM
const jokeSection = JokeSection();
const joke = Joke();

root.append(jokeSection);
jokeSection.append(joke);
```

I needed some time to understand what's happening here, and I guess the coach has to explain it during the session as well. I see that the fetched data is passed in line 32 `joke.append(randomJoke.joke);` to this prepared `joke`.

What I don't know is the best practice here:
Would it make more sense to write a function `createJoke(joke)` where all `creteElement` calls are done; we could call it in line 32 where it takes the `randomeJoke.joke` as argument? By doing so, we would not distract the students from what's happening during the fetch.

But this is only a suggestion, your way might be better!

sessions/js-fetch/demo-end/js/index.js
```diff
@@ -0,0 +1,40 @@
...
+			console.error('Bad Response');
+		} else {
+			// Success
+			const jokeData = await response.json();
+
+			const randomIndex = getRandomInt(0, jokeData.length);
+
+			const randomJoke = jokeData[randomIndex];
+
+			joke.append(randomJoke.joke);
```
djfarly, line 32:
```suggestion
			joke.textContent = randomJoke.joke;
```

sessions/js-fetch/demo-end/js/index.js
```diff
@@ -0,0 +1,40 @@
...
+		const response = await fetch('https://example-apis.vercel.app/api/bad-jokes');
+
+		// Failure (Bad response)
+		if (!response.ok) {
+			console.error('Bad Response');
+		} else {
+			// Success
+			const jokeData = await response.json();
+
+			const randomIndex = getRandomInt(0, jokeData.length);
```
djfarly, line 28:
```suggestion
			const randomIndex = getRandomInt(0, jokeData.length - 1);
```

PReichetanz (approved):

PR #122, opened 2022-10-26
JS Unit Testing: getAnimal Challenge

PReichetanz:
rendered: https://github.com/neuefische/web-exercises/tree/js-unit-testing-get-animal-challenge/sessions/js-unit-testing/get-animal

## Challenge to review
- To review this challenge locally, use the following command in your terminal:
`npx ghcd@latest neuefische/web-exercises/tree/js-unit-testing-get-animal-challenge/sessions/js-unit-testing/get-animal -i`

refers to neuefische/web-curriculum-new-format#160


sessions/js-unit-testing/get-animal/index.test.js
```diff
@@ -0,0 +1,14 @@
+import {getAnimal} from './index.js';
+
+test('getAnimals() returns a string "I really like cats!" if called with the argument "cat".', () => {
+	const result = getAnimal('cat');
+	expect(result).toBe('I really like cats!');
+});
+test('getAnimals() returns a string "I like [animalPluralName]!" if called with a string argument.', () => {
+	const result = getAnimal('dog');
+	expect(result).toBe('I like dogs!');
```
aclaes, line 9: While working on this exercise my mind was boggling about building the plural form of an animal name. It's easy with "dog", but simple adding an "s" at the end wouldn't work with "goose". This might be an additional hurdle for students. 

I think we could overcome this issue by passing a plural form into the function (`getAnimal("dogs")`); Student would only need to write the logic to handle "undefined" and "cats" different from other inputs.

sessions/js-unit-testing/get-animal/index.test.js
```diff
@@ -0,0 +1,14 @@
+import {getAnimal} from './index.js';
+
+test('getAnimals() returns a string "I really like cats!" if called with the argument "cat".', () => {
+	const result = getAnimal('cat');
+	expect(result).toBe('I really like cats!');
```
aclaes, line 5: The difference between "I really like cats!" and "I like dogs!" is quite subtile.

It might be a good idea to change the sentence so it is more clear to students they need to handle the input `"cat"` different than other inputs. Idea:

> "I totally love cats!"

aclaes (changes requested):
Good challenge! I have some minor improvement ideas

sessions/js-unit-testing/get-animal/index.test.js
PReichetanz: I totally see your point here.

If we only use plural forms, there is one disadvantage: 
Inside the `getAnimal` function, students only have to check the prop to equal 'cats' or 'dogs', but they don't need to use the prop in the return statement as in `return <p>I totally like {animal}s.</p>`

If we can live with that and want the students to focus on understanding unit tests and write a simpler function, I am fine with that approach.

aclaes (approved):
🚀 approve

PR #123, opened 2022-10-26
JS Unit Testing: add demo

aclaes:
test local with

```
npx ghcd@latest neuefische/web-exercises/tree/js-unit-testing-demo/sessions/js-unit-testing/demo-start 
```


djfarly:
Paul remarks should be fixed with the update

PReichetanz:
> Paul remarks should be fixed with the update

I checked the demo again and the issues are fixed, thank you!

PR #128, opened 2022-10-27
migrate recap project 2 and 3

aclaes:


sessions/recap-project-3/rick-and-morty-app/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "rick-and-morty-starter",
```
djfarly, line 2: currently ghcd will use the package name as the folder name it creates

We should do one of those:
1. rename the package to `rick-and-morty-app`
2. or add another param to ghcd alá `ghcd <github-path> <foldername>` that would be consistent with how clone and other tools work

now that I think of it… why not both…

sessions/recap-project-2/quiz-app/README.md
```diff
@@ -0,0 +1,90 @@
+# Quiz App
```
djfarly, line 1: recap project 2 does not have a package json so i'll use this random line to comment…

When ghcd does not find a package json it tries to mash together a folder name and it will become something like:

web-exercises-quiz-app

We can of course change that behaviour to just use the dirname.

We could also add a simple package.json that does nothing besides having a name.

Or implement the addition to ghcd described above.

djfarly (changes requested):
We should think about how ghcd will clone the projects:

sessions/recap-project-3/rick-and-morty-app/package.json
djfarly:
```suggestion
  "name": "rick-and-morty-app",
```

sessions/recap-project-2/quiz-app/README.md
aclaes: Oh, good point! 

I added a `package.json` file. 

From the first HTML basic exercises on there is a `package.json`file. Students should be used to it, even if they don't know the purpose yet.

sessions/recap-project-3/rick-and-morty-app/package.json
aclaes: I like the idea of having a `ghcd` argument for a folder name. I put an issue into the backlog.

https://github.com/neuefische/web-curriculum-new-format/issues/243

PR #134, opened 2022-11-01
Fix broken image link

CJAkkuec:
Small fix regarding the "expected result" image in the place-elements challenge. :)


aclaes (approved):
🚀

PR #157, opened 2022-11-11
fix(js-array-methods): refactor second test of map challenge

PReichetanz:
## Reasoning
- Students needed to much time figuring out the usage of single and double quotes to make the second test work.
- It's not necessary to use both kinds of quotes in this test in order to understand how the `map` method works.

## Changes
- change the expectation of `questionsAndAnswersTogether` to expect only one string containing the question and answer, separated by a dash `-`.

fixes neuefische/web-curriculum-new-format#247

### DIFF
--- a/sessions/js-array-methods/map/index.test.js
+++ b/sessions/js-array-methods/map/index.test.js
@@ -14,9 +14,9 @@ describe("Mapping Challenge", () => {
   });
   test("questionsAndAnswersTogether", () => {
     expect(questionsAndAnswersTogether).toEqual([
-      '"How often can I use <header>?" - "As often as you like."',
-      '"How often can I use <aside>?" - "As often as you like."',
-      '"On which types can I use destructuring assignment?" - "On Objects and Arrays"',
+      "How often can I use <header>? - As often as you like.",
+      "How often can I use <aside>? - As often as you like.",
+      "On which types can I use destructuring assignment? - On Objects and Arrays",
     ]);
   });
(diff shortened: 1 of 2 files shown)


sessions/js-array-methods/map/index.test.js
```diff
@@ -14,9 +14,9 @@ describe("Mapping Challenge", () => {
   });
   test("questionsAndAnswersTogether", () => {
     expect(questionsAndAnswersTogether).toEqual([
-      '"How often can I use <header>?" - "As often as you like."',
-      '"How often can I use <aside>?" - "As often as you like."',
-      '"On which types can I use destructuring assignment?" - "On Objects and Arrays"',
+      "How often can I use <header>? - As often as you like.",
+      "How often can I use <aside>? - As often as you like.",
+      "On which types can I use destructuring assignment? - On Objects and Arrays",
```
djfarly, line 19: 👍 
this needs to be changed here as well:
https://github.com/neuefische/web-exercises/blob/2bbd0c7b17a666e925215fb251976cfcbf7a0dad/sessions/js-array-methods/map/index.js#L29

djfarly (changes requested):
works! see comment below :)

sessions/js-array-methods/map/index.test.js
PReichetanz: True, forgot about that, thank you!

djfarly (approved):
🥰

PR #185, opened 2022-11-22
Js functions pizza analyzer

F-Kirchhoff:
### 🔎 Review

Review via [🔗 **CodeSandbox**](https://codesandbox.io/s/github/neuefische/web-exercises/tree/js-functions-pizza-analyzer/sessions/js-functions/pizza-analyzer?file=/README.md) or locally by running this command in your Terminal:

```
npx ghcd@latest neuefische/web-exercises/tree/js-functions-pizza-analyzer/sessions/js-functions/pizza-analyzer -i
```

````md
Solve this challenge on
[🔗 **CodeSandbox**](https://codesandbox.io/s/github/neuefische/web-exercises/tree/main/sessions/js-functions/pizza-analyzer?file=/README.md)
or locally by running this command in your Terminal:

```
npx ghcd@latest neuefische/web-exercises/tree/main/sessions/js-functions/pizza-analyzer -i
```
````

### DIFF
--- a/sessions/js-functions/pizza-analyzer/.eslintrc.json
+++ b/sessions/js-functions/pizza-analyzer/.eslintrc.json
@@ -1,0 +1,11 @@
+{
+  "env": {
+    "browser": true,
+    "es2022": true,
+    "node": true
+  },
+  "extends": ["eslint:recommended"],
+  "parserOptions": {
+    "sourceType": "module"
+  }
+}
--- a/sessions/js-functions/pizza-analyzer/README.md
+++ b/sessions/js-functions/pizza-analyzer/README.md
@@ -1,0 +1,44 @@
+# js-functions: Pizza Analyzer
+
+Pizza is amazing. So we want the most pizza for our money. Lets build a tool to compare two pizza sizes and gives us the net plus in pizza area.
+
+## Task
+
+<!--
+
+Explaining the task in detail. E.g.:
+
+Use JavaScript to react to the form submission.
+
+Log all form data (in object form) into the console in the submit event handler.
+
+You can use the following hints as guideline:
+
+- Hint 1
+- Hint 2
+- ...
+
+Switch to the  [`js/index.js`](./js/index.js)  file and make something great happen!
+
+-->
+
+## Notes
+
+- You only have to touch the [`js/index.js`](./js/index.js) file.
+
+## Development
+
+### CodeSandbox
+
+Select the "Browser" tab to view this project.
+
+### Local development
+
+To run project commands locally, you need to install the dependencies using `npm i` first.
+
+You can then use the following command:
+
+- `npm run lint` to run the linter
+
+Use the Live Preview Extension for Visual Studio Code to view this project in the browser.  
+Select the HTML file you want to view, press <kbd>⇧</kbd><kbd>⌘</kbd><kbd>P</kbd>, search for `Live Preview: Show Preview` and confirm with <kbd>Enter</kbd>.
--- a/sessions/js-functions/pizza-analyzer/README.md
+++ b/sessions/js-functions/pizza-analyzer/README.md
@@ -4,23 +4,35 @@ Pizza is amazing. So we want the most pizza for our money. Lets build a tool to 
 
 ## Task
 
-<!--
+### Calculate the Pizza Gain
 
-Explaining the task in detail. E.g.:
+First, we want to calculate the difference of the two pizza areas for two given pizza diameters. For that, do the following:
 
-Use JavaScript to react to the form submission.
+1. Write a function `calculatePizzaGain` which expects the two diameters of the two pizzas: `diameter1` and `diameter2`. (Not the radii of the pizzas!)
+2. Inside this function, calculate the areas of both pizzas. (Hint: you can find the formula for that online)
+3. Calculate the pizza gain in percent relative to the first pizza (Hint: the formula is: `(area2 - area1) / area1 * 100)`
+4. Set the `textContent` of the output element to that value.
+5. Inside both event listeners, save the current values of both pizza inputs in two variables `pizzaSize1` and `pizzaSize2` (Hint: you can use the queried elements and get their value like this: `pizzaInput.value`)
+6. Now call the function and pass in the two pizza sizes as the arguments.
 
-Log all form data (in object form) into the console in the submit event handler.
+### Change pizza dipslay sizes
 
-You can use the following hints as guideline:
+Great! Now we know how much pizza we gain if we choose the second pizza. But we don't have any visual relations between these two numbers. So lets adapt the pizza displays.
 
-- Hint 1
-- Hint 2
-- ...
+1. Write a function `updatePizzaDisplay` which has two parameters: `pizzaElement` - the queried DOM element of the respective pizza and `newSize` which is the new size of that pizza.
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

File: sessions/js-functions/pizza-analyzer/README.md [hunk 1]
```diff
@@ -0,0 +1,56 @@
...
+Select the "Browser" tab to view this project.
+
+### Local development
+
+To run project commands locally, you need to install the dependencies using `npm i` first.
+
+You can then use the following command:
+
+- `npm run lint` to run the linter
+
```
shebtastic, line 54: are the students already familiar with linting here?

sessions/js-functions/pizza-analyzer/js/index.js
```diff
@@ -0,0 +1,29 @@
...
+pizzaInput2.addEventListener("input", () => {
+  // write your code here
+});
+
+function calculatePizzaGain(diameter1, diameter2) {
+  // write your code here
+}
+
+function updatePizzaDisplay(pizzaElement, newSize) {
+  // write your code here
```
shebtastic, line 24: should we comment that this function belongs to task 2? and the others respectively?

sessions/js-functions/pizza-analyzer/index.html
```diff
@@ -0,0 +1,52 @@
...
+      name="viewport"
+      content="width=device-width, initial-scale=1, shrink-to-fit=no"
+    />
+    <title>js-functions: Pizza Analyzer</title>
+    <link rel="stylesheet" href="./css/styles.css" />
+    <script src="./js/index.js" defer></script>
+  </head>
+  <body>
+    <main>
+      <h1>Pizza Aanalyzer 🍕</h1>
```
shebtastic, line 15:
```suggestion
      <h1>Pizza Analyzer 🍕</h1>
```

sessions/js-functions/pizza-analyzer/.eslintrc.json
```diff
@@ -0,0 +1,11 @@
+{
+  "env": {
+    "browser": true,
+    "es2022": true,
+    "node": true
+  },
+  "extends": ["eslint:recommended"],
```
shebtastic, line 7: should this include `"prettier"`?

sessions/js-functions/pizza-analyzer/README.md
```diff
@@ -0,0 +1,56 @@
...
+Pizza is amazing. So we want the most pizza for our money. Lets build a tool to compare two pizza sizes and gives us the net plus in pizza area.
+
+## Task
+
+### Calculate the Pizza Gain
+
+First, we want to calculate the difference of the two pizza areas for two given pizza diameters. For that, do the following:
+
+1. Write a function `calculatePizzaGain` which expects the two diameters of the two pizzas: `diameter1` and `diameter2`. (Not the radii of the pizzas!)
+2. Inside this function, calculate the areas of both pizzas. (Hint: you can find the formula for that online)
```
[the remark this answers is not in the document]
F-Kirchhoff: Ah true, that might be useful!

sessions/js-functions/pizza-analyzer/README.md
F-Kirchhoff: oh thats from the template I will remove it.

sessions/js-functions/pizza-analyzer/.eslintrc.json
F-Kirchhoff: I am not sure, I used the template from the new format team. I guess it should be discussed with them.

djfarly: not nessesary unless you want to use eslint as a formatter :)

sessions/js-functions/pizza-analyzer/README.md
```diff
@@ -0,0 +1,55 @@
...
+1. Write a function `updatePizzaDisplay` which has two parameters: `pizzaElement` - the queried DOM element of the respective pizza and `newSize` which is the new size of that pizza.
+2. Calculate the new display size of the pizza element
+   > 💡 The formula is: `newSize / 24 \* 100`.
+3. Set the width of the `pizzaElement` to this new value.
+   > 💡 you can use `pizzaElement.style.width` for that.
+4. Call this function in both event listeners and pass in the correct pizza element and the correct pizza size.
+
+### 3 Extra: Change the Output Background
+
+Lets update the background of the output section to indicate whether we gain or loose pizza if we choose pizza 2.
```
PReichetanz, line 34:
```suggestion
Let's update the background of the output section to indicate whether we gain or loose pizza if we choose pizza 2.
```

sessions/js-functions/pizza-analyzer/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+1. Write a function `updateOutputColor` with two parameters: `size1` and `size`.
```
PReichetanz, line 36:
```suggestion
1. Write a function `updateOutputColor` with two parameters: `size1` and `size2`.
```

PReichetanz (commented):
This is a nice and exciting exercise with a lot of real-world importance! ❤️

Please see [my solution here](https://codesandbox.io/s/js-functions-pizza-analyzer-solution-01-l0b6l1?file=/js/index.js), but I'm not able to set the correct new width after clicking the buttons, so my pizzas don't change; what am I missing here?

I also tried `calculatedDisplaySize.toString()`, but it didn't work either.

PReichetanz:
> This is a nice and exciting exercise with a lot of real-world importance! ❤️
> 
> Please see [my solution here](https://codesandbox.io/s/js-functions-pizza-analyzer-solution-01-l0b6l1?file=/js/index.js), but I'm not able to set the correct new width after clicking the buttons, so my pizzas don't change; what am I missing here?
> 
> I also tried `calculatedDisplaySize.toString()`, but it didn't work either.

I think the changes after review work fine! 🚀

Just one note about my issue quoted above:
To be honest, setting CSS styles via JavaScript is not my expertise, so I would say this problem and the not working solution is totally on me.
But anyway, do the students, while solving this challenge, know about how to set the width dynamically because it was part of (let's say) a session two days ago?

If so, everything's fine here; if not, we might add a hint how to do this (because setting a style attribute is not the focus of this challenge, so they should not use too much time to find out how to do this, I would say).

F-Kirchhoff:
@PReichetanz: Thats a good point there. For changing the width there is a hint, but I could add that they need to add a "px" at the end. For the colors, I could also create two classes which do the trick. What do you think?

PReichetanz:
> @PReichetanz: Thats a good point there. For changing the width there is a hint, but I could add that they need to add a "px" at the end. For the colors, I could also create two classes which do the trick. What do you think?

The hint itself is ok, but I didn't know that I have to add "px" at the end, so students might struggle here as well?

💡 Maybe it's just me because I just had to write markdown files with `` elements where you don't add the "px" part when setting the width.

I think that the explanation for the colors is fine and we don't have to change it (because students don't need to add anything else when setting `outputSection.style.backgroundColor = "var(--red)";`.

F-Kirchhoff:
I am pretty sure the students will struggle with it. So we should make it easier for them. I will think about it.

PReichetanz (approved):
LGTM!

PR #190, opened 2022-11-24
add recipes, starters and new next.js recipe

djfarly:


README.md
```diff
@@ -27,18 +33,48 @@ All of the exercises are designed to work locally as well as on CodeSandbox.
 
 ## Structure of this Repository
 
-A _exercise_ is any form of exercise, challenge, demo or starter that is used in the web bootcamps.
+### Sessions and Exercises
 
 A _session_ is a collection of exercises that are used in a specific bootcamp session.
 
-The folder structure of this repository is as follows:
+A _exercise_ is any form of exercise, challenge, demo or starter that is used in the web bootcamps.
```
PReichetanz, line 40:
```suggestion
An _exercise_ is any form of exercise, challenge, demo or starter that is used in the web bootcamps.
```

recipes/README.md
```diff
@@ -0,0 +1,54 @@
+# 🍝 Recipes
+
+This directory contains recipes that can be used to create templates and starters.
+
+## Templates vs. Starters
+
+- **Templates** are used to create **new exercises, challenges, etc. directly in this repository** using the `npm run create` command.
+- **Starters** are used to create **new projects outside of this repository** using the `npx ghcd neuefische/web-exercises/starters/<starter> my-app -i` command.
+
+The main difference content-wise is that templates `README.md` file contains an instructions template for the exercise, while starters `README.md` file is very bare-bones. See [How?](#how) for more details.
```
PReichetanz, line 10:
```suggestion
The main difference content-wise is in regards to the `README.md` file: in templates, it contains an instructions template for the exercise, while in starters, it is very bare-bones. See [How?](#how) for more details.
```

recipes/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+> **Warning**  
+> Never edit the files in the `starters` and `templates` directories directly. Instead, edit the recipe in this folder and then [cook it using the `npm run cook <recipe>` command](#cooking-a-recipe).
+
+## Why?
+
+Basic templates and starters only contain files. But we want some more advanced variants for Create React App and Next.js that are created via automation. The `cook` script automates the creation of all templates and starters from recipes.
```
PReichetanz, line 17: This is a little bit confusing to me:
- In contrast to basic templates and starters, do Create React App and Next.js templates / starters contain something else than files?
- This is why it is not so clear to me **why exactly** we are in need of such a recipe: because we want to have a single source of truth (as soon as there is a version update, etc.) so that we can simply update all templates + starters automatically?

recipes/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+## How?
```
PReichetanz, line 19: This part is quite clear, I think.

Regarding the "why" again: sounds like the only reason is that the `readme.md` is different in templates and starters (and other files). We should not repeat other reasons (if there are some), but I think we should be a little bit more clear about our "why" in the respective section above.

PReichetanz (commented):
I checked the `main readme.md` and the `recipes/readme.md` files.

The difference between templates and starters is clear, the overview about all available templates / starters is done well.

The only issue in `recipes/readme` belongs to the "why": I think we can be more specific here that this makes it easier to maintain the templates / starters as soon as there are version updates (if this is a reason, I'm not sure).

By doing so, we can avoid the question "all this effort just because of the different readme files?"

recipes/cra/files/sandbox.config.json
```diff
@@ -0,0 +1 @@
+{ "template": "create-react-app" }
```
djfarly, line 1:
```suggestion
{ "template": "next" }
```

aclaes (changes requested):
Looks very good 🚀 I tested the cooking process for all recipes and the creationg of execercises from all cooked templates. Works nice!

Found just some small issues

aclaes (approved):
Look very good and works rock solid 🪨

PR #199, opened 2022-11-30
remove condescending language

djfarly:
I noticed we sometimes use just or simple which can seem condescending:

"This is just state" does not feel good if state is an unknown concept.

It can be removed without any problems, so here we go:


aclaes (approved):
LGTM

PR #209, opened 2022-12-05
Replace nanoid with uid

PReichetanz:
Replace all instances of `nanoid` with `uid`.

There are some instances left in the `package-lock.json`.

### DIFF
--- a/.ncurc.json
+++ b/.ncurc.json
@@ -1,4 +1,4 @@
 {
   "upgrade": true,
-  "deep": true,
+  "deep": true
 }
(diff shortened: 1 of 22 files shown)


.ncurc.json
```diff
@@ -1,5 +1,5 @@
 {
   "upgrade": true,
   "deep": true,
-  "reject": ["nanoid"]
+  "reject": ["uid"]
```
djfarly, line 4: we dont need this anymore - it was only here to prevent `ncu` from updating nanoid automatically

```suggestion
```

djfarly (approved):
i have spot checked it and it seems to work

PR #215, opened 2022-12-07
removed timoutId logic and replaced with disabled button

doemser:


PReichetanz (commented):
I see the advantage of the start button being disabled as long as the toast is visible. Additionally, I like that we don't redeclare the `timeoutId` to `undefined` inside of the `hideToast` function.

I think it's fine, but I'm not an expert on timeouts and animations, so I'm not sure whether this is more charming.

ThomasF85 (approved):
Looks good and easier!

PR #223, opened 2022-12-12
React Data Fetching: Star Wars API Challenge

PReichetanz:
### 🔎 Review

Review locally by running this command in your Terminal:

```
npx ghcd@latest neuefische/web-exercises/tree/react-data-fetching_swapi/sessions/react-data-fetching/star-wars -i
```

````md
Solve this challenge locally by running this command in your Terminal:

```
npx ghcd@latest neuefische/web-exercises/tree/main/sessions/react-data-fetching/star-wars -i
```
````

refers to neuefische/web-curriculum-new-format#216

### CODE REVIEW

File: sessions/react-data-fetching/star-wars/pages/characters/luke-skywalker.js [hunk 1]
```diff
@@ -0,0 +1,15 @@
+import Card from "../../components/Card";
+import Layout from "../../components/Layout";
+
+export default function Character() {
```
aclaes, line 4: The transition from the static to the dynamic route might be a bit hard to understand. Maybe we can prepare somthing.

- rename file to `pages/characters/1.js`
- add `id` as a constant that needs toe be replaces with a query param via `useRouter`: `const id = 1;`
- pass and render the  `id` to the `Card` so it does make some sense in the beginning: `<Card id={id} ... />`

sessions/react-data-fetching/star-wars/components/Card/index.js
```diff
@@ -0,0 +1,16 @@
+export default function Card({ name, height, eyeColor, birthYear }) {
+  return (
+    <div>
```
aclaes, line 3: We could use a styled component adding a border or a background here to make it appear like a card ;-)

sessions/react-data-fetching/star-wars/pages/index.js
```diff
@@ -0,0 +1,11 @@
+import Link from "next/link";
+import Layout from "../components/Layout";
+
+export default function HomePage() {
+  return (
+    <Layout>
+      <h1>React Data Fetching: Star Wars</h1>
+      <Link href="characters/luke-skywalker">Link to first character</Link>
```
aclaes, line 8: How do you feel about adding some links to characters to the homepage? 

This way students easier understand what to do and can quickly test the dynamic routes.

We should mention in the README that only the first link works in the beginning.

```
      <ul>
        <li>
            <Link href="/characters/1">Luke Skywalker</Link>
        </li>
        <li>
            <Link href="/characters/2">C-3PO</Link>
        </li>
        <li>
            <Link href="/characters/3">R2-D2</Link>
        </li>
        <li>
            <Link href="/characters/4">Darth Vader</Link>
        </li>
      </ul>
```

aclaes (changes requested):
Would we like this to be the first or the second challange? If it is the first, we should implement the logic with plain `fetch` and have the students adopt it to `SWR`. For a second challenge this is fine!

I think we should adjust the setup slightly to make thr transition to dynamic routes easier.

sessions/react-data-fetching/star-wars/pages/characters/luke-skywalker.js
PReichetanz: Good idea, changed it this way!

sessions/react-data-fetching/star-wars/components/Card/index.js
```diff
@@ -0,0 +1,16 @@
...
+      <h2>Name: {name}</h2>
```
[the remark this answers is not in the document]
PReichetanz: True for the headline; we don't need the info headline here and it makes quite some sense to delete it.

aclaes (approved):
Looks very good. Approve 🚀

PR #242, opened 2023-01-04
feat: add demo start and end to css-positioning exercises

zartinn:
Does it make sense to add missing demos when we prepare for lessons?

### DIFF
--- a/sessions/css-positioning/demo-end/css/styles.css
+++ b/sessions/css-positioning/demo-end/css/styles.css
@@ -1,0 +1,78 @@
+*,
+*::before,
+*::after {
+  box-sizing: border-box;
+}
+
+body {
+  display: flex;
+  flex-direction: column;
+  padding: 0;
+  margin: 0;
+  background: papayawhip;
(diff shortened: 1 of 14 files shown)


sessions/css-positioning/demo-end/css/styles.css
```diff
@@ -0,0 +1,78 @@
...
+
+.disclaimer {
+  margin: 20px 5%;
+  margin-top: 150px;
+  border: 2px solid black;
+  padding: 1em;
+}
+
+.disclaimer__title {
+  left: 55px;
```
F-Kirchhoff, line 42: I guess the position attribute is missing here.

sessions/css-positioning/demo-end/css/styles.css
```diff
@@ -0,0 +1,78 @@
...
+}
+
+.disclaimer__icon {
```
F-Kirchhoff, line 45: here as well

F-Kirchhoff (changes requested):
in the "end" version you missed the position attributes, otherwise it looks good to me.

F-Kirchhoff (approved):
LGTM

PR #243, opened 2023-01-04
Backend Create Demo

CJAkkuec:
## Copyable markdown for `web-curriculum-new-format`:
Use this demo locally by running this command in your Terminal:

```
npx ghcd@latest neuefische/web-exercises/tree/main/sessions/backend-create/demo-start -i
```

Use this demo locally by running this command in your Terminal:

```
npx ghcd@latest neuefische/web-exercises/tree/main/sessions/backend-create/demo-end -i
```


sessions/backend-create/demo-end/pages/api/jokes/[id].js
```diff
@@ -0,0 +1,17 @@
...
+
+  if (request.method === "GET") {
+    const joke = await Joke.findById(id);
+
+    if (!joke) {
+      return response.status(404).json({ status: "Not Found" });
+    }
+
+    response.status(200).json(joke);
+  }
```
Comment (pixelass) [inline sessions/backend-create/demo-end/pages/api/jokes/[id].js:16]: else case is missing and should respond with a 501 `response.status(501).json({error: "not implemented"})` or `response.status(501).end()`

sessions/backend-create/demo-end/pages/api/jokes/index.js
```diff
@@ -0,0 +1,24 @@
...
+      const jokeData = request.body;
+      const joke = new Joke(jokeData);
+      await joke.save();
+
+      response.status(201).json({ status: "Joke created" });
+    } catch (error) {
+      console.log(error);
+      response.status(400).json({ error: error.message });
+    }
+  }
```
pixelass, line 23: see comment above about 501 response

sessions/backend-create/demo-start/db/connect.js
```diff
@@ -0,0 +1,47 @@
+import mongoose from "mongoose";
```
pixelass, line 1: see comment in  end for the new version of this file

pixelass (approved):
please review the open suggestions and comments

sessions/backend-create/demo-end/db/connect.js
```diff
@@ -0,0 +1,47 @@
+import mongoose from "mongoose";
```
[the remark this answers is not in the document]
CJAkkuec: Hey there :) I've investigated a little bit, and it seems that your suggestion uses MongoClient instead of mongoose? I tried to research whether or not the two are interchangeable, but it appears that they work differently, while mongoose is using MongoClient "under the hood". Since we've used mongoose so far, I'd be a little hesitant to change it? The [db/connent.js ](https://github.com/vercel/next.js/blob/canary/examples/with-mongodb-mongoose/lib/dbConnect.js)here utilizing mongoose also seems to be the one that we're using currently. I'd appreciate your insight. :)

sessions/backend-create/demo-end/next.config.js
```diff
@@ -0,0 +1,18 @@
...
+  reactStrictMode: true,
+  webpack(config) {
+    config.module.rules.push({
+      test: /\.svg$/i,
+      issuer: /\.[jt]sx?$/,
+      use: ["@svgr/webpack"],
+    });
+
+    return config;
+  },
```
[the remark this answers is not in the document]
CJAkkuec: Do you suppose that this would be a generally applicable change? If so, I'd like to open an issue for it and adjust it everywhere. :)

sessions/backend-create/demo-end/pages/api/jokes/[id].js
Comment (CJAkkuec) [reply sessions/backend-create/demo-end/pages/api/jokes/[id].js:16]: I'd like to implement this in every other instance as well, hence why I'd like to open an issue for this as well. :)

sessions/backend-create/demo-end/db/connect.js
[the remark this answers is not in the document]
pixelass: true true, my bad. please ignore

sessions/backend-create/demo-end/next.config.js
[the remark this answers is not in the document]
pixelass: I think it is an overhead if we never import any SVG (probably the default case). I suggest only adding additional config if it is actually needed.

pixelass:
I reviewed your response and added comments where applicable. 
In general I agree and think the adjustments can be done in later steps. Please open the proposed issues, so we can track the improvement.

sessions/backend-create/demo-end/components/JokeForm/index.js
```diff
@@ -0,0 +1,39 @@
+import useSWR from "swr";
+
+export default function JokeForm() {
+  const jokes = useSWR("/api/jokes");
+
+  async function onSubmit(event) {
```
PReichetanz, line 6:
```suggestion
  async function handleSubmit(event) {
```

sessions/backend-create/demo-end/components/JokeForm/index.js
```diff
@@ -0,0 +1,39 @@
...
+      await response.json();
+      jokes.mutate();
+      event.target.reset();
+    } else {
+      console.error(`Error: ${response.status}`);
+    }
+  }
+
+  return (
+    <form id="joke-form" onSubmit={onSubmit}>
```
PReichetanz, line 33:
```suggestion
    <form id="joke-form" onSubmit={handleSubmit}>
```

PR #246, opened 2023-01-06
Add missing .env.local to root .gitignore file

PReichetanz:
## Rationale
When I create a challenge which uses a MongoDB, I create a `.env.local` file in the corresponding project (let's say `/sessions/backend-create/products`.

While working in that branch locally, everything is fine because the project has its own `.gitignore` file.

Unfortunately, as soon as I switch to a branch where `/sessions/backend-create/products` does not yet exist (e.g. in `main`), there are some leftovers: a hidden `.next` folder, the `node_modules` folder and my recently created `.env.local`.

The problem: there is now no `.gitignore` file which is why `.env.local` is tracked by git.

See this screenshot:
Screenshot 2023-01-06 at 14 51 51

## Solution
I add '.env.local' to the root `.gitignore`.

## Question
Is there a better way to handle this or can anything break by doing so?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,9 +47,8 @@ node_modules/
 *.tgz
 
 # dotenv environment variables file
-.env
-.env.local
-.env.test
+.env*
+!.env.local.example
 
 # parcel-bundler cache (https://parceljs.org/)
 .cache
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,7 @@ node_modules/
 
 # dotenv environment variables file
 .env
+.env.local
 .env.test
 
 # parcel-bundler cache (https://parceljs.org/)


.gitignore
```diff
@@ -47,8 +47,8 @@ node_modules/
 *.tgz
 
 # dotenv environment variables file
-.env
-.env.test
+.env*
+!.env.local.example
```
bluepostit, line 51: Looks good to me :) 

Is there any possibility of others wanting to add an env file with some other name pattern? Eg. `.env.example`?

PReichetanz: I don't think so: we only use `.env` files in the context of Next.js which wants us to use a `.env.local` for our secrets.

So as far as I can see, there is no need for a `.env.example` right now.

badabam: There is a chance, that this might happen at some point. The solution for me would be for that person to add another line `!.env.example` which would be an explicit and secure way in my opinion.

badabam (approved):
We discussed this. It's a good solution in my opinion.

PR #247, opened 2023-01-06
update swr dependency

ThomasF85:
If the students follow the current SWR docs example [here](https://swr.vercel.app/#overview), they will run into an error with SWR version < 2.0.0, since the current docs do not contain a check for `!data`, but instead only check for `error` and `isLoading`. You can only do this with SWR version >= 2.0.0 => hence the dependency update.


PReichetanz (approved):
Thanks, @ThomasF85!

I've checked the entire `web-exercises` repository: this is the only place where the version of `swr` is >2.0.0, so I think this is fine!

PR #250, opened 2023-01-10
Backend: Update and Delete Demo

JessicaLoers:
### 🔎 Review

Review locally by running this command in your Terminal:

```
npx ghcd@latest neuefische/web-exercises/tree/backend-update-delete_demo/sessions/backend-update-delete/demo-start -i
```

```
npx ghcd@latest neuefische/web-exercises/tree/backend-update-delete_demo/sessions/backend-update-delete/demo-end -i
```

````md
Use this demo locally by running this command in your Terminal:

```
npx ghcd@latest neuefische/web-exercises/tree/main/sessions/backend-update-delete/demo-start -i
```
```
npx ghcd@latest neuefische/web-exercises/tree/main/sessions/backend-update-delete/demo-end -i
```
````

### CODE REVIEW

File: sessions/backend-update-delete/demo-start/pages/_app.js [hunk 1]
```diff
@@ -0,0 +1,15 @@
+import { SWRConfig } from "swr";
+import GlobalStyle from "../styles";
+
+const fetcher = (url) => fetch(url).then((r) => r.json());
```
F-Kirchhoff, line 4: I would use "response" or at least "res" here.

sessions/backend-update-delete/demo-start/pages/api/jokes/[id].js
```diff
@@ -0,0 +1,17 @@
+import dbConnect from "../../../db/connect";
+import Joke from "../../../db/models/Joke";
+
+export default async function handler(request, response) {
+  await dbConnect();
+  const { id } = request.query;
+
+  if (request.method === "GET") {
```
Comment (F-Kirchhoff) [inline sessions/backend-update-delete/demo-start/pages/api/jokes/[id].js:8]: if the request method is not "GET" here, you will get a timeout, the other cases should be handled as well with a send of a status 405.

sessions/backend-update-delete/demo-end/components/Button/index.js
```diff
@@ -0,0 +1,20 @@
...
+  box-shadow: 0px 1px 5px -2px var(--color-granite);
+`;
+
+export default function Button({ width, margin, children, type, onClick }) {
+  return (
+    <StyledButton type={type} width={width} margin={margin} onClick={onClick}>
+      {children}
+    </StyledButton>
+  );
+}
```
F-Kirchhoff, line 20: Is this component necessary? I guess you can use the StyledButton, because the Button does not contribute any functionality or renaming of props.

sessions/backend-update-delete/demo-end/components/JokeForm/index.js
```diff
@@ -0,0 +1,37 @@
+import styled from "styled-components";
+import Button from "../Button";
+import { useState } from "react";
+
+const StyledForm = styled.form`
+  display: flex;
+  flex-direction: column;
+  gap: 1rem;
+  margin-block-start: 2rem;
```
F-Kirchhoff, line 9: nice! (just learned about this in the josh comeau Kurs)

sessions/backend-update-delete/demo-end/components/JokeList/index.js
```diff
@@ -0,0 +1,30 @@
...
+import useSWR from "swr";
+import { useRouter } from "next/router";
+
+export default function JokeList() {
+  const router = useRouter();
+  const { data } = useSWR("/api/jokes");
+
+  if (!data) {
+    return <h1>Loading...</h1>;
+  }
```
F-Kirchhoff, line 13: shouldn't we use the isLoading from useSWR here?

sessions/backend-update-delete/demo-end/components/JokeList/index.js
```diff
@@ -0,0 +1,30 @@
...
+  return (
+    <List>
+      {data.map((joke) => (
+        <ListItem key={joke._id}>
+          <Button
+            type="button"
+            width="100%"
+            onClick={() => router.push(`/${joke._id}`)}
+          >
+            {joke.joke}
```
F-Kirchhoff, line 24: I am not sure but I think I remember that we wanted to use the Link component from Next.js here.

sessions/backend-update-delete/demo-end/components/List/index.js
```diff
@@ -0,0 +1,13 @@
...
+  list-style-type: none;
+  display: grid;
+  gap: 1rem;
+  justify-items: center;
+  padding: 0;
+`;
+
+export default function List({ children }) {
+  return <StyledList>{children}</StyledList>;
+}
```
F-Kirchhoff, line 13: This component can be omitted. You can export the styledList directly.

sessions/backend-update-delete/demo-end/components/ListItem/index.js
```diff
@@ -0,0 +1,9 @@
+import styled from "styled-components";
+
+const StyledListItem = styled.li`
+  width: 100%;
+`;
+
+export default function ListItem({ children }) {
+  return <StyledListItem>{children}</StyledListItem>;
+}
```
F-Kirchhoff, line 9: Can be omitted as well, if the StyledListItem is exported instead.

sessions/backend-update-delete/demo-end/components/ListItem/index.js
```diff
@@ -0,0 +1,9 @@
+import styled from "styled-components";
+
+const StyledListItem = styled.li`
+  width: 100%;
```
F-Kirchhoff, line 4: I am not sure but do we need the with 100% here? i thought list item are block elements.

sessions/backend-update-delete/demo-end/pages/[id].js
```diff
@@ -0,0 +1,59 @@
...
+    push("/");
+  }
+
+  async function handleDeleteJoke() {
+    const response = await fetch(`/api/jokes/${id}`, {
+      method: "DELETE",
+    });
+
+    if (response.ok) {
+      await response.json();
```
Comment (F-Kirchhoff) [inline sessions/backend-update-delete/demo-end/pages/[id].js:47]: this is not used anywhere else (the json), I guess this can be omitted.

sessions/backend-update-delete/demo-end/pages/api/jokes/[id].js
```diff
@@ -0,0 +1,29 @@
...
+    if (!joke) {
+      return response.status(404).json({ status: "Not Found" });
+    }
+
+    response.status(200).json(joke);
+  }
+
+  if (request.method === "PUT") {
+    const jokeToUpdate = await Joke.findByIdAndUpdate(id, {
+      $set: request.body,
```
Comment (F-Kirchhoff) [inline sessions/backend-update-delete/demo-end/pages/api/jokes/[id].js:20]: I think you need to parse the body first: `JSON.parse(request.body)`

F-Kirchhoff (changes requested):
This is looking very good to me, but I found some unnecessary code that can be removed. Have a look into the comments!

sessions/backend-update-delete/demo-start/pages/_app.js
CJAkkuec: Absolutely, an issue to fix this already exists. Thanks! :)

sessions/backend-update-delete/demo-start/pages/api/jokes/[id].js
Comment (CJAkkuec) [reply sessions/backend-update-delete/demo-start/pages/api/jokes/[id].js:8]: Nice catch! :) For this, too, an issue already exists. (Totally understandable that you're unaware.) We'll try to get to this asap!

sessions/backend-update-delete/demo-end/components/Button/index.js
CJAkkuec: It's not strictly necessary, no. Seeing as how this is a demo that the students will be confronted with during the lecture part of the session, I figured it'd be best to keep any styling as far away as possible and the code as logic-focused as possible. Does that make sense?

sessions/backend-update-delete/demo-end/components/JokeForm/index.js
CJAkkuec: Modern CSS (@container, @layer, :has, :is, etc ...) is extremely fun and useful - I love it sm.

sessions/backend-update-delete/demo-end/components/JokeList/index.js
CJAkkuec: We absolutely should.

sessions/backend-update-delete/demo-end/components/JokeList/index.js
CJAkkuec: @JessicaLoers let me know that we've also created an issue for this. If you're wondering why we're creating so many issues: it's a cascade of problems. This is the final stage of four consecutive sessions, so we have to address stuff that has been repeating and compounding once everything's done structurally. :)

sessions/backend-update-delete/demo-end/components/ListItem/index.js
CJAkkuec: I'll test it!

sessions/backend-update-delete/demo-end/pages/api/jokes/[id].js
Comment (JessicaLoers) [reply sessions/backend-update-delete/demo-end/pages/api/jokes/[id].js:20]: This  made me curious, so I tested it.  🕵️‍♀️  

Because we use the `$set` expression we don't have to parse the body. 
An alternative solution could be:

```javascript
  if (request.method === "PUT") {
    const data = JSON.parse(request.body);
    const jokeToUpdate = await Joke.findByIdAndUpdate(id, data);
    return response.status(200).json(jokeToUpdate);
  }
```
I can't judge nor decide which approach is the better 🫡 right 🤗 nicer 😎 cooler one. Do you?

sessions/backend-update-delete/demo-end/components/List/index.js
JessicaLoers: Yes, right 🧹

sessions/backend-update-delete/demo-end/components/Button/index.js
F-Kirchhoff: Not really, I would teach the students to write their code as simple and understandable as possible, adding an extra component layer that does nothing but renaming the component locally in the file introduces more complexity and more stuff to read/parse/understand. I would argue lets remove this extra component, the StyledButton should be kept in this extra file of course. So you have the separation of styling and functionality but not two components in one file where one is obsolete.

sessions/backend-update-delete/demo-end/pages/api/jokes/[id].js
Comment (F-Kirchhoff) [reply sessions/backend-update-delete/demo-end/pages/api/jokes/[id].js:20]: Interesting! thats actually quite useful.

Comment (F-Kirchhoff) [reply sessions/backend-update-delete/demo-end/pages/api/jokes/[id].js:20]: Hmm, I don't see much difference in complexity. cant decide

sessions/backend-update-delete/demo-end/components/JokeForm/index.js
F-Kirchhoff: maybe you can give a short workshop about that? I have no idea what you are talking about :D

sessions/backend-update-delete/demo-end/components/Button/index.js
CJAkkuec: Seeing as how this is a demo, the students shouldn't even be shown the content of the styled components. 🤷‍♀️ I'll try to find a solution.

sessions/backend-update-delete/demo-end/components/JokeForm/index.js
CJAkkuec: Nah 😂

sessions/backend-update-delete/demo-end/pages/api/jokes/[id].js
Comment (CJAkkuec) [reply sessions/backend-update-delete/demo-end/pages/api/jokes/[id].js:20]: I'll decide :D We'll keep it as is for the time being, as it was a recommendation by former employees.

sessions/backend-update-delete/demo-end/components/ListItem/index.js
CJAkkuec: So, this is behavior specific to `<li>` apparetly, because it's neither `display: block`, nor `display: inline-block`, it's `display: list-item`, which apparently shares some default behavior with inline-block elements. 🤷‍♀️ Wild.

sessions/backend-update-delete/demo-start/components/Button/index.js
```diff
@@ -0,0 +1,20 @@
...
+  box-shadow: 0px 1px 5px -2px var(--color-granite);
+`;
+
+export default function Button({ width, margin, children, type, onClick }) {
+  return (
+    <StyledButton type={type} width={width} margin={margin} onClick={onClick}>
+      {children}
+    </StyledButton>
+  );
+}
```
F-Kirchhoff, line 20:
```suggestion
export default styled.button`
  width: ${({ width }) => width};
  margin: ${({ margin }) => margin};
  appearance: none;
  border: none;
  background: var(--color-water-10);
  font-size: larger;
  padding: 0.5rem 1rem;
  box-shadow: 0px 1px 5px -2px var(--color-granite);
`;
```

F-Kirchhoff (changes requested):

sessions/backend-update-delete/demo-start/components/Button/index.js
CJAkkuec: Thank you for your suggestion, I've already implemented this! :) (Pretty much exactly like this, only as a named export.)

F-Kirchhoff: It seems that it is not yet pushed? Because I can't find it in the commits

JessicaLoers: Don't worry, we are still working on the demo and the session guide. The changes will follow soon - we promise. 🤞

F-Kirchhoff (approved):
LGTM 🦊

PR #251, opened 2023-01-17
fix: add body to index.html

zartinn:


zartinn:
in general the tasks in the css are not always 100% clear what to select actually select. I may propose a different task in a couple of days.

F-Kirchhoff (approved):
LGTM

PR #256, opened 2023-01-27
Update: improve instructions for some challenges

bluepostit:
Improved the wording of a few of the `js-loops` challenges, adding better clarity. This is based on grammatical rules as well as on feedback from students.

### DIFF
--- a/sessions/js-loops/forin-select-nav/README.md
+++ b/sessions/js-loops/forin-select-nav/README.md
@@ -24,9 +24,9 @@ Before we create a navigation bar, let's have a look what's already there in [`j
 - Each value is an object containing an "href" and "text" key and a string value.
 - A `nav` and `ul` element is created and appended to the `main` element.
 
-Your taks is to fill this list with list items and anchor tags; as content, use the `nav` object which you can access with a `for...in` loop.
+Your task is to fill this list with list items and anchor tags; as content, use the `nav` object which you can access with a `for...in` loop.
 
-Use the following hints as guideline:
+Use the following hints as a guideline:
 
 - Write a `for...in` loop for the `nav` object.
 - In each iteration,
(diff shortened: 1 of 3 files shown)


sessions/js-loops/forin-select-nav/README.md
```diff
@@ -44,9 +44,7 @@ Select the "Browser" tab to view this project.
 
 ### Local development
 
-To run project commands locally, you need to install the dependencies using `npm i` first.
-
-You can then use the following command:
+You can check your code for problems with the following command:
```
doemser, line 47: why removed the info to `npm i`?

sessions/js-loops/forof-list/README.md
```diff
@@ -26,9 +26,7 @@ Select the "Browser" tab to view this project.
 
 ### Local development
 
-To run project commands locally, you need to install the dependencies using `npm i` first.
-
-You can then use the following command:
+You can check your code for problems with the following command:
```
doemser, line 29: same here

doemser (commented):
Good improvements, but why give up the `npm i` hint? Am I missing something?

PReichetanz:
> [...] Also removed some irrelevant instructions from README files: no need to tell the students to install dependencies with `npm i`, when there are no dependencies listed in the `package.json`.

I see the point. The reason for this unnecessary `npm i` is the [`migrate.js` script](https://github.com/neuefische/web-exercises/blob/main/scripts/migrate.js): it makes sure that all challenges derived from the same template looks the same.

Every time we'll change the general structure of a template (in the case of the challenges involved here, the `"js"` template), we should run the migrate script again to ensure consistency.

The problem is the following: if we decide to change another template, we need to run the `migrate` script as well, but this will reset the changes suggested by this PR.

As far as I understand this, deleting these lines will be reset until the run the `migrate` script again because we have changed a template.

I think we should improve the `migrate` script instead: make sure that the `npm i` command is not part of the `README.md` file if there are no dependencies.

bluepostit:
Thanks, @doemser and @PReichetanz. Sounds like an issue to be solved elsewhere. For now, I'll remove those changes.

PReichetanz (approved):
LGTM!

F-Kirchhoff (approved):
LGTM 🦊

PR #257, opened 2023-01-27
Fix/backend read exercise

onemanwenttomow:
* fixes some minor typos.

* removes the finished schemas that the students are instructed to create.


PReichetanz (approved):
Nice find, thank you for the PR, I have nothing to add!

PR #259, opened 2023-02-02
Css Selectors Demo

ahohnsen:
Updated demo:

- I deleted the `:root` selector including all CSS variables and used color names instead 👉🏼 The `:root` selector is actually part of the css-structure session, so I would keep this demo a bit more simple
- I updated the explanation for the adjacent sibling selector

**❓ Open questions/suggestions:**

- The combinators are part of the demo but not of the session guide. I suggest to add a short section in the session guide for combinators if we teach them during this session.
- The session guide shows different code examples for the selectors than the demo. I would propose to rewrite the session guide, so you can follow along the demo with the guide (as for example in the js-conditions-and-booleans session.

Let me know, what you think 🙂  

### 🔎 Review

Review via [🔗 **CodeSandbox**](https://codesandbox.io/s/github/neuefische/web-exercises/tree/css-selectors-demo/sessions/css-selectors/demo-start?file=/README.md) or locally by running this command in your Terminal:

```
npx ghcd@latest neuefische/web-exercises/tree/css-selectors-demo/sessions/css-selectors/demo-start -i
```

### 🔎 Review

Review via [🔗 **CodeSandbox**](https://codesandbox.io/s/github/neuefische/web-exercises/tree/css-selectors-demo/sessions/css-selectors/demo-end?file=/README.md) or locally by running this command in your Terminal:

```
npx ghcd@latest neuefische/web-exercises/tree/css-selectors-demo/sessions/css-selectors/demo-end -i
```

### DIFF
--- a/sessions/css-selectors/demo-end/css/styles.css
+++ b/sessions/css-selectors/demo-end/css/styles.css
@@ -35,7 +35,7 @@ of an HTML tag. Like <section class="main-section">. */
 
 .main-section {
   padding: 20px;
-  background-color: mintcream;
+  background-color: lavenderblush;
   margin-bottom: 20px;
 }
 
@@ -63,7 +63,6 @@ ul[role="list"] {
 a[href^="https"] {
   color: hotpink;
   padding: 8px;
-  margin: 0 0 8px 0;
 }
 
 /* PART 4: I'm a pseudo-class. I'm  a keyword added to a selector that 
--- a/sessions/css-selectors/demo-end/css/styles.css
+++ b/sessions/css-selectors/demo-end/css/styles.css
@@ -1,19 +1,3 @@
-/* I am the root pseudo-class. The html type selector and
-I target pretty much the same stuff, but my specificity is 
-higher. I'm frequently used to store global CSS variables. */
-
-:root {
-  /* Color Variables */
-
-  --color-nemo: #ff4a11;
-  --color-granite: #252629;
-  --color-water-10: #f3f5f9;
-  --color-foam: #ffffff;
-
-  /* Unit Variables */
-  --gap: 20px;
-}
-
 /* I'm the universal selector. I target EVERYTHING. */
 *,
 *::before,
@@ -29,17 +13,17 @@ body {
 }
 
 h1 {
-  color: var(--color-nemo);
-  text-shadow: 3px 3px var(--color-granite);
+  color: orangered;
+  text-shadow: 3px 3px darkslategray;
 }
 
 h1,
 h2 {
-  margin: var(--gap);
+  margin: 20px;
 }
 
 h3 {
-  margin-bottom: var(--gap);
+  margin-bottom: 20px;
 }
 
 a {
@@ -54,9 +38,9 @@ of an HTML tag. Like <section class="main-section">. */
 }
 
 .main-section {
-  padding: var(--gap);
-  background-color: var(--color-water-10);
-  margin-bottom: var(--gap);
+  padding: 20px;
+  background-color: mintcream;
+  margin-bottom: 20px;
 }
 
 .box {
@@ -73,7 +57,7 @@ of an HTML tag. Like <section class="main-section">. */
 as part of HTML tags. */
 
 button[type="button"] {
-  padding: var(--gap);
+  padding: 20px;
 }
 
 ul[role="list"] {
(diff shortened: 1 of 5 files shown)


sessions/css-selectors/demo-end/css/styles.css
```diff
@@ -29,17 +13,13 @@ body {
 }
 
 h1 {
-  color: var(--color-nemo);
-  text-shadow: 3px 3px var(--color-granite);
+  color: orangered;
+  text-shadow: 3px 3px darkslategray;
```
CJAkkuec, line 17:
```suggestion
  color: #ff4a11;
  text-shadow: 3px 3px #252629;
```
Please use the neuefische colors :). They are the same for every challenge. I don't see how the `:root` selector is an issue here, because it's not part of the demonstration. I'm not dying on this hill, though. 🤷‍♀️

sessions/css-selectors/demo-end/css/styles.css
```diff
@@ -54,9 +34,9 @@ of an HTML tag. Like <section class="main-section">. */
 }
 
 .main-section {
-  padding: var(--gap);
-  background-color: var(--color-water-10);
-  margin-bottom: var(--gap);
+  padding: 20px;
+  background-color: lavenderblush;
```
CJAkkuec, line 38: Strictly speaking, the general consensus is to use hex colors instead of color names. I don't really mind either way. But since we're using hex codes for the neuefische colors, we should probably keep using hex codes.

sessions/css-selectors/demo-end/css/styles.css
```diff
@@ -122,7 +101,7 @@ a[href^="https"]:before {
 of the parent element, but only those descendants that match my type. */
 
 .main-section .text {
-  color: var(--color-granite);
+  color: darkslategray;
```
CJAkkuec, line 104:
```suggestion
  color: #252629;
```

sessions/css-selectors/demo-start/css/styles.css
```diff
@@ -29,17 +13,14 @@ body {
 }
 
 h1 {
-  color: var(--color-nemo);
-  text-shadow: 3px 3px var(--color-granite);
+  color: orangered;
+  text-shadow: 3px 3px darkslategray;
```
CJAkkuec, line 17:
```suggestion
 color: #ff4a11;
  text-shadow: 3px 3px #252629;
```

sessions/recap-project-1/quiz-app/README.md
```diff
@@ -18,6 +18,7 @@ Start by:
 
 - Creating a new folder for your project.
 - Initialize a local git repository.
+- Create a .gitignore file and add .DS_Store to the file.
```
CJAkkuec, line 21:
```suggestion
- Create a `.gitignore` file and add `.DS_Store` to the file.
```

CJAkkuec (changes requested):

sessions/css-selectors/demo-end/css/styles.css
ahohnsen: Ok, then hex colors. In the previous demo it was a mix of hex colors and color names. 🙂

sessions/css-selectors/demo-end/css/styles.css
ahohnsen: 👍🏼 Ah, that makes sense. I did not notice that they are used everywhere :) 
Yes, you are right, it is already part of the start version of the demo. But when adding styles during the demonstration we will use the CSS variables for colors and margin/padding. In previous sessions I had the feeling that this confuses the students. 

In general, my point is to either introduce the `:root` selector in this session or as part of the css-structure session. So we do not have to explain the same thing twice 🤷🏼‍♀️

PReichetanz: I would also argue to introduce the `:root` selector in the CSS Structure session when talking about variables. I'm afraid that this would confuse students too much at this point.

PReichetanz:
>❓ Open questions/suggestions:

  >  The combinators are part of the demo but not of the session guide. I suggest to add a short section in the session guide for combinators if we teach them during this session.

Yes, I agree that the session guide should mirror the demo and help coaches to teach the session.

>    The session guide shows different code examples for the selectors than the demo. I would propose to rewrite the session guide, so you can follow along the demo with the guide (as for example in the js-conditions-and-booleans session.

This should be changed as well. Should we create an issue to tackle this?

ahohnsen:
> > ❓ Open questions/suggestions:
> 
> > The combinators are part of the demo but not of the session guide. I suggest to add a short section in the session guide for combinators if we teach them during this session.
> 
> Yes, I agree that the session guide should mirror the demo and help coaches to teach the session.
> 
> > The session guide shows different code examples for the selectors than the demo. I would propose to rewrite the session guide, so you can follow along the demo with the guide (as for example in the js-conditions-and-booleans session.
> 
> This should be changed as well. Should we create an issue to tackle this?

Yes, sounds good! 🙌🏼

CJAkkuec (approved):
LGTM ♥️

PR #261, opened 2023-02-03
CSS Basics

actyralf:
- css-basics/pokemon-page: update task #6 to use left/right borders instead of underline to prevent solutions with text-decoration:underline
- css-basics/social-card: update demo image to reflect the same arrow style as used in the SVG


CJAkkuec:
LGTM ;)

PR #264, opened 2023-02-16
(Rework)CSS: Positioning Demo

CJAkkuec:
Rework demo-start and demo-end, corresponds with this [PR](https://github.com/neuefische/web-curriculum-new-format/pull/332). (Here's a [link](https://github.com/neuefische/web-curriculum/pull/332) that should work after the repo name change.)

### DIFF
--- a/sessions/css-positioning/demo-end/index.html
+++ b/sessions/css-positioning/demo-end/index.html
@@ -11,32 +11,32 @@
   </head>
   <body>
     <header class="header">
-      <h1 class="header__heading">
+      <h1 class="header__headline">
         Sweeties<span class="plate" role="img" aria-label="A plate">🍽</span>
       </h1>
     </header>
 
-    <main class="main">
-      <section class="main__section">
-        <p class="main__text">
+    <main class="content">
+      <section class="content__section">
+        <p class="content__text">
           From hard candies to gummies, chocolates to caramels, we've got
           something for everyone to satisfy their sweet tooth!
         </p>
         <span class="donut" role="img" aria-label="A chocolate donut">🍩</span>
         <span class="donut-text">Delicious!</span>
-        <p class="main__text">
+        <p class="content__text">
           First, our hard candies come in a variety of flavors, including
           classic favorites like cherry, lemon, and lime, as well as unique
           blends like watermelon-mint and strawberry-basil. They're perfect for
           enjoying slowly, letting the flavor melt and linger in your mouth.
         </p>
-        <p class="main__text">
+        <p class="content__text">
           For those who prefer a chewier candy, our gummies are a must-try.
           We've got everything from classic gummy bears to sour worms and fruit
           slices, with flavors like grapefruit, peach, and raspberry. Our
           gummies are soft and chewy, bursting with fruity goodness.
         </p>
-        <p class="main__text">
+        <p class="content__text">
           For the chocolate lovers out there, Sweeties has you covered with our
           delicious chocolate creations. From rich, creamy truffles to
           chocolate-covered pretzels, there's a treat to suit every taste. We
@@ -46,15 +46,15 @@
         <span class="separator" role="img" aria-label="A piece of hard candy"
           >🍬</span
         >
-        <p class="main__text">
+        <p class="content__text">
           Our caramels are some of the best around. We've got classic flavors
           like vanilla and chocolate, as well as unique twists like sea salt and
           honey. Our caramels are soft and buttery, with just the right amount
           of sweetness.
         </p>
-        <p class="main__text info">
+        <p class="content__text content__text--info">
           <span
-            class="info-accent"
+            class="content__accent"
             role="img"
             aria-label="A piece of hard candy"
             >🍭</span
@@ -64,7 +64,7 @@
           for a sweet treat for yourself or a gift for someone special, our
           range of candies is sure to please.
         </p>
-        <p class="main__text">
+        <p class="content__text">
           In addition to our classic candies, Sweeties also offers a range of
           seasonal and special occasion treats. For example, during the winter
           holidays, we create festive candy canes in flavors like peppermint,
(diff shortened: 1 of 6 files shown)


sessions/css-positioning/demo-start/index.html
```diff
@@ -6,216 +6,119 @@
...
-          Lorem ipsum dolor sit, amet consectetur adipisicing elit. Alias vitae
-          libero harum! Odit consequatur veritatis, hic qui, facere distincLorem
-          ipsum dolor sit, veritatis, hic qui, facere distincorem ipsum dolor
-          sit, amet consectetur adipisicing elit. Alias vitae libero harum! Odit
-          consequatur veritatis, hic qui, facere distinc
+
+    <main class="main">
+      <section class="main__section">
+        <p class="main__text">
+          From hard candies to gummies, chocolates to caramels, we've got
```
MerleMak, line 22: 😂👍🏼

MerleMak (approved):
Looks great!

sessions/css-positioning/demo-end/index.html
```diff
@@ -6,215 +6,119 @@
...
       content="width=device-width, initial-scale=1, shrink-to-fit=no"
     />
-    <title>HTML & CSS</title>
+    <title>CSS Positioning</title>
     <link rel="stylesheet" href="./css/styles.css" />
   </head>
   <body>
     <header class="header">
-      <h1 class="header__heading">CSS positioning</h1>
+      <h1 class="header__heading">
```
MerleMak, line 14:
```suggestion
      <h1 class="header__headline">
```
sounds a little nicer in my ears, but not a priority!

sessions/css-positioning/demo-end/index.html
```diff
@@ -6,215 +6,119 @@
...
-        <h2 class="disclaimer__title">disclaimer</h2>
-        <div class="disclaimer__icon">!</div>
-        <p class="disclaimer__text">
-          Lorem ipsum dolor sit, amet consectetur adipisicing elit. Alias vitae
-          libero harum! Odit consequatur veritatis, hic qui, facere distincLorem
-          ipsum dolor sit, veritatis, hic qui, facere distincorem ipsum dolor
-          sit, amet consectetur adipisicing elit. Alias vitae libero harum! Odit
-          consequatur veritatis, hic qui, facere distinc
+
+    <main class="main">
```
MerleMak, line 19:
```suggestion
    <main class="content">
```
Maybe a more descriptive class name helps to understand better why BEM is useful.

sessions/css-positioning/demo-end/index.html
```diff
@@ -6,215 +6,119 @@
...
+      <section class="main__section">
+        <p class="main__text">
+          From hard candies to gummies, chocolates to caramels, we've got
+          something for everyone to satisfy their sweet tooth!
+        </p>
+        <span class="donut" role="img" aria-label="A chocolate donut">🍩</span>
+        <span class="donut-text">Delicious!</span>
+        <p class="main__text">
```
MerleMak, line 27:
```suggestion
        <p class="content__text">
```

sessions/css-positioning/demo-end/index.html
```diff
@@ -6,215 +6,119 @@
...
+        <span class="separator" role="img" aria-label="A piece of hard candy"
+          >🍬</span
+        >
+        <p class="main__text">
+          Our caramels are some of the best around. We've got classic flavors
+          like vanilla and chocolate, as well as unique twists like sea salt and
+          honey. Our caramels are soft and buttery, with just the right amount
+          of sweetness.
+        </p>
+        <p class="main__text info">
```
MerleMak, line 55:
```suggestion
        <p class="content__text content__info">
```

sessions/css-positioning/demo-end/index.html
```diff
@@ -6,215 +6,119 @@
...
+          <span
+            class="info-accent"
```
MerleMak, line 57:
```suggestion
            class="content__info--accent"
```

MerleMak (commented):
Some BEM comments, but nothing big :)

sessions/css-positioning/demo-end/index.html
CJAkkuec: The little icon here isn't a variant of `content__info` 🤔 .

sessions/css-positioning/demo-end/index.html
CJAkkuec: I'm not sure if that's how BEM works. It could be argued that it should be `content__text content__text--info`, because it's a variant of `content__text`, but that's really up in the air imho.

sessions/css-positioning/demo-end/index.html
CJAkkuec: I'm indifferent, if you think this should be changed, I'll change it everywhere. :)

sessions/css-positioning/demo-end/index.html
CJAkkuec: See above. :)

sessions/css-positioning/demo-end/index.html
MerleMak: You are right, I think content__text--info is the way to go. Imo the point of BEM is to be able to make sense of what and where the defined class is used for, just by the help of the class name. Just "info" doesn't really do that, because if you just see this class it could be located and used anywhere. In this case I think that is not the purpose of this info.

sessions/css-positioning/demo-end/index.html
MerleMak: I feel like "content" is a bit more descriptive about the Block.
I don't know if you know this little trick,I just learned this during my Research time, so i want to share: If you right-click on the class name in the HTML file and click on "Rename Symbol" every actually related (equal) name will be renamed. Its not renaming the string, which is super useful :)

sessions/css-positioning/demo-end/index.html
MerleMak: You are right, I was distracted by the info part. 
info__accent would be the way to go I guess. one `-` is not sufficient either way :)

MerleMak: BEM is hard.. I don't actually know if you have to stay consistent with your Blocks, or if you can redefine the Blocks now and then. Because content__text--info makes sense to me in the larger context, but here the info you be the new block and referring to the same thing

PR #265, opened 2023-02-23
Feature/js callback functions demo with button component

F-Kirchhoff:
Add a specific real life example for creating a higher order function - a button component, which accepts an onClick callback function


JessicaLoers (approved):
LGTM 🚀

PR #269, opened 2023-03-10
change Position of the hint to be better found

Roland-Hufnagel:
the hint for exercise 1 should be placed after exercise 1 (not after exercise 2). So I changed it's position.


doemser (approved):
Yes! :)

mpagels (approved):

PR #274, opened 2023-03-21
fix: make sure the active class is toggled on the button 

ThomasKirst:
The active class must be toggled on the button element(event.currentTarget)
instead of the svg or path element (current.target) (which could have been the trigger for the event)

https://user-images.githubusercontent.com/4458383/226622565-86cf4649-71fe-439e-a754-b762cdb53635.mov


PReichetanz (approved):
Looks good to me, thank you for pointing this out!

PR #275, opened 2023-03-23
Feat/js fetch color clue

NikRoe:
replace "the closest named" color with "the random" color for more clarity


PReichetanz (approved):
Sounds good to me, seems to be more clear now.

PR #276, opened 2023-03-24
change test text to match function name

NikRoe:
changed text in test description to match the function name that is imported and supposed to be used here


PReichetanz (approved):
LGTM!

PR #277, opened 2023-03-27
fix: remove function declarations

F-Kirchhoff:
Since the students should practice writing function declarations and the spiced students got confused with the task descriptions, I removed the declarations from the starter code.


PReichetanz (approved):
I see why we are doing this, although I can imagine that our students might be confused how to handle parameters etc. But let's try it this way.

F-Kirchhoff:
You are right we should be aware about this possible pitfall. They should be able to research this though in the handout.

PR #280, opened 2023-04-04
Update Recap Project 2: Add note for form page

mbosselmann:
In Recap Project 2, we found out that students don't think about the fact that they can also create multiple JavaScript files, and include a different JavaScript file for each page. However, if students get as far as adding JavaScript for their form during this day, they run into the problem of event listeners that throw error messages for DOM elements that does not exist for the form page.

Therefore, I suggest adding this as a note and am open to discussion and further suggestions. :)


F-Kirchhoff (approved):
LGTM 🦊

PR #281, opened 2023-04-05
Update space rocket challenge

mbosselmann:
**Problem:** Locally the task could not be solved because `global` was used. If the sequence was started when pressing Start, there was an error message: "global is not defined".

Together with @shebtastic we rebuilt the Space Rocket Challenge so that it would now work locally if the challenge was solved correctly and also the tests will pass.

This solution should not close [this issue](https://github.com/neuefische/web-curriculum/issues/315), but can be a solution until this issue is finally discussed.

I tested everything before creating the PR, but feel free to try it remotely and locally to see if it works for you. As a solution for the `launchSequence.js` to see if the rocket starts without error message, you can copy this one in:

```js
// Implement the launch sequence function here and export it as the default export.
import { loadPayload } from "./core/load.js";
import { NFSAT, FISHSAT } from "./payload/satellites.js";
import { fuel } from "./core/fuel.js";
import { countdown } from "./core/countdown.js";
import { liftoff } from "./core/liftoff.js";
import { deployPayload } from "./core/deploy.js";
// Implement the launch sequence function here and export it as the default export.

export default function launch() {
  loadPayload(NFSAT);
  loadPayload(FISHSAT);
  fuel();
  countdown(5);
  countdown(4);
  countdown(3);
  countdown(2);
  countdown(1);
  liftoff();
  deployPayload();
}
```

### DIFF
--- a/sessions/js-structure/spacerocket/jest.config.js
+++ b/sessions/js-structure/spacerocket/jest.config.js
@@ -1,0 +1,3 @@
+export default {
+  testEnvironment: "jsdom",
+};
--- a/sessions/js-structure/spacerocket/js/core/rocket.js
+++ b/sessions/js-structure/spacerocket/js/core/rocket.js
@@ -17,9 +17,9 @@ export function getNewRocket() {
 }
 
 export function getRocket() {
-  if (!global.rocket) {
+  if (!window.rocket) {
     throw new Error("No rocket found");
   }
 
-  return global.rocket;
+  return window.rocket;
 }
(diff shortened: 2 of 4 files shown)


F-Kirchhoff:
Thank you for your PR, but I agree with Dominik. For now the exercise is moved to a "Bonus" Section of the session, but if we all agree on it in a team meeting, we can remove the challenge completely.

mbosselmann:
That is true, but currently we hand out this challenge and it is not possible to solve it locally (but we do not mention this and provide a link to get a local version). As long as the decision to remove this challenge is not finalized we can use this and this works fine and no student is frustrated anymore, because his solution does not work locally (like I described above).

F-Kirchhoff (approved):
LGTM

PR #283, opened 2023-04-06
change next-version to 13.2.4 in React-Styled-Components

Roland-Hufnagel:
To better use next.js font optimization I changed the next-version in the package.json to 13.2.4 in the session "React-Styled-Components"

### DIFF
--- a/sessions/react-styled-components/box/package.json
+++ b/sessions/react-styled-components/box/package.json
@@ -12,11 +12,10 @@
   "dependencies": {
     "eslint": "8.29.0",
     "eslint-config-next": "13.0.6",
-    "next": "13.0.6",
+    "next": "13.2.4",
     "react": "18.2.0",
     "react-dom": "18.2.0",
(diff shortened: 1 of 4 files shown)


doemser:
Looks good. The web-exercises repository uses a script to create templates. I suspect that your change would be overwritten if you were to re-cook the templates at some point:

https://github.com/neuefische/web-exercises/blob/main/recipes/README.md

Look over there and evaluate if the customization needs to be done in the scripts as well.

It would also be interesting to know how you get along with the cook script, or if that makes editing and fixing errors unnecessarily difficult and does not outweigh the advantage.

djfarly:
The cook script only overwrites the templates, so you should be fine just updating the exercise itself.

NikRoe:
So my conclusion from your comments is:

it looks good and the cook script should not overwrite those changes.

--> it can be merged

NikRoe (approved):

PR #284, opened 2023-04-06
delete @ from next/font/google-path

Roland-Hufnagel:
after updating next from 13.0.6 to 13.2.4 the character '@' is deprecated in:
import { Open_Sans } from "@next/font/google";


F-Kirchhoff (approved):
LGTM

PR #287, opened 2023-04-17
update next template to next 13.3

djfarly:


recipes/next/files/pages/_document.js
```diff
@@ -22,4 +22,16 @@ export default class MyDocument extends Document {
       sheet.seal();
```
djfarly, line 22: seal

doemser:
Why don't we use the src directory? Students tend to create a src folder that doesn't contain the whole source like "pages", we could prevent that easily.

Besides this, there was a new process established for changes like this which is documented in the web-curriculum README:

https://github.com/neuefische/web-curriculum

doemser:
As I understand the new process, changing the template is from now on a change that has to be decided in the team. Please follow that process and create an issue for this.

We implemented the process for this kind of decisions and you are right that we need to update our next template.

djfarly:
Any larger change to this template or how it is used may be discussed in the team.
This is just a version bump, according to Felix this is a patch, thus i'll merge this now.

PR #288, opened 2023-04-24
Fix/backend links

F-Kirchhoff:
For all Demos of `backend-api-routes`,`backend-read`,`backend-create` and `backend-update-and-delete`:

- replace all buttons for redirecting with next/link components
- use isLoading in all demos
- update styling

closes https://github.com/neuefische/web-curriculum/issues/292


PReichetanz (approved):
LGTM!

PR #291, opened 2023-04-25
Minor Change: Update CSS Selectors Demo

mbosselmann:
- as part of the update of the CSS Selectors guide I want to propose two small changes for the CSS Selectors Demo:
  - remove *::before and *::after: pseudo elements will be explained later during the session, but these stylings are there from the start. This could be confusing and I think it is not necessary for this session that we work with ::before and ::after from the beginning
  - I would not remove default margins and paddings with the universal selector as it may be better to have some space between the examples from the beginning
  - and I would propose to remove the last combinator example of the demo. Two examples should be enough (as the session is already long enough) and they can learn more about combinators with the handout


F-Kirchhoff (approved):
LGTM

PR #292, opened 2023-04-26
refactor: use links instead of buttons and add isLoading condition

F-Kirchhoff:
- use next/links instead of buttons
- add isLoading condition
- update backend-update-and-delete products challenge starter code to use useSWRMutation


mbosselmann (approved):
LGTM! :octocat:

PR #295, opened 2023-05-04
Refactor: React Component Testing - Scorekeeper Challenge

PReichetanz:
## Rationale
My experience has shown that students are quite overwhelmed when starting the challenge. Especially the second sub task - fixing the bug in the `HistoryEntry` component - is confusing. We don't really show fixing bugs during the session, so starting with this task seems to be problematic.

## Suggestion
I moved the sub task regarding the `HistoryEntry` component to the end of the challenge. Students still start with discussing which features should be tested, and then write their first, simple test case for the `Navigation`. Following the rest of the challenge, the tasks grow in their complexity.

### DIFF
--- a/sessions/react-component-testing/scorekeeper/README.md
+++ b/sessions/react-component-testing/scorekeeper/README.md
@@ -27,20 +27,7 @@ The app offers a test file for every component. Each of these files contain acce
 
 Your task is to run `npm run test`, write the test implementations and make the tests pass.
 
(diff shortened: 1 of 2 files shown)


F-Kirchhoff:
This is a nice improvement. Especially the usage of test.skip. By this the students can focus on the first challenge and only later on start fixing the bugs.

F-Kirchhoff (approved):

PR #309, opened 2023-08-11
chore: reverted changes from before when git was planned to be introduced after the quiz app setup

ThomasKirst:


ThomasF85 (approved):
Looks good!

PR #317, opened 2023-08-30
Fix: create-react-app related warning concerning babel dependency

ThomasKirst:
This PR fixes the warning shown in React app which had been create using CRA.

Screenshot 2023-08-23 at 11 54 15

You can try it out with this command, for example:

```
npx ghcd@latest neuefische/web-exercises/tree/fix-cra-warning-babel-dependency/sessions/react-state-2/colored-number -i
```

Run `npm i` and `npm start`. The warning should be gone!

Screenshot 2023-08-30 at 12 02 51

### DIFF
--- a/recipes/README.md
+++ b/recipes/README.md
@@ -13,13 +13,13 @@ They can also be used to create new **projects outside of this repository** usin
 
 ## Why?
 
-Basic templates only contain files. But we want to do more advanced things for Create React App and Next.js: Those templates are created using `create-react-app` and `create-next-app` and then we modify them to our needs by adding and modifiying files as well as running additional commands.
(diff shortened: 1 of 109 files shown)


onemanwenttomow (approved):
Warnings gone for me 🎉

I do get the following when i run `npm i`

``` bash
6 high severity vulnerabilities 
```

but you get that when you use the latest version of CRA (so i don't think there is anything we can do about it?)

`npx create-react-app my-app`

ThomasKirst:
Hi Pete, thanks for testing it!

> 6 high severity vulnerabilities 
Yes, that´s unfortunate, but we cannot really do anything about it since the CRA development is stalled ://

PR #339, opened 2023-11-24
Fix minor inconsistency with naming styled components

ThomasKirst:
This PR resolves two minor issues:
- The README tells students they would only have to touch the `index.js` whereas they will have to work in the `_app.js`
- All styled components should be prefixed with `Styled` as per our convention.


mbosselmann (approved):
LGTM! :octocat:

PR #344, opened 2024-01-03
Update/react basics

MerleMak:


sessions/react-basics/demo-start/src/index.js
```diff
@@ -0,0 +1,20 @@
...
+// This is the dom element that React will render into,
+// per convention, it is usually a div with an id of "root".
+const reactRootElement = document.querySelector("#root");
+
+// This is the "react" way to create a "root" that react can render into.
+const reactRoot = createRoot(reactRootElement);
+
+// Call ".render()" on the root to render the elements into the DOM.
+// Notice that this is NOT a template literal, it is JSX.
+reactRoot.render();
```
F-Kirchhoff, line 20: I would suggest to write this code together with the students, then they have time to digest what each part is actually doing.
```suggestion
```

F-Kirchhoff (changes requested):
Looks good! I only would write the createRoot and root.render() code live with the students. What do you think?

sessions/react-basics/demo-start/src/index.js
doemser: @F-Kirchhoff 
I never wrote it myself when starting a project.

F-Kirchhoff: ok good to know, I myself showed it every time. In my opinion this is just a small detail. @MerleMak, what do you think about it?

doemser: I was not talking about showing it to students, but about the need to write it yourself at any point in the future.. 

This code is always given by tool-chains/bundlers/frameworks like create-react-app, vite or next.js.. most developers will never need to write it, which makes it an unnecessary payload for the react basics intro.

F-Kirchhoff: I can see both sides here. It is true that this will add extra content to the to the session, altough writing 2 lines of extra code is not that much. Most of the students will be fine with a quick explaination about these lines of code, but I have seen students that were really grateful for building this code up from the ground, especially when they need to understand the fundamentals. 

I personally don't know which way is more appropriate and in the end it will not really matter that much, since it is a small detail. I wanted to highlight this other possibility, but I would let @MerleMak decide after hearing both sides.

F-Kirchhoff (approved):
I change my review to an approval. @MerleMak should merge or update this PR.

PR #359, opened 2024-02-26
change naming of variable in the challenge

Roland-Hufnagel:
since the variable name: 'name' leads to this warining:
```js 'name' is deprecated.ts(6385)
lib.dom.d.ts(27240, 5): The declaration was marked as deprecated here.
```

I have changed it to 'userName'.


marcbendixen (approved):
lgtm 🧑‍🚀

PR #360, opened 2024-02-26
reworked the challenge

Roland-Hufnagel:
The problem was, that each time, you checked an expression using the console, the form was reset and the radio buttons were no longer selected. 
I reworked the challenge to have a button that shows if your answers have been right or wrong. 
And I added two more expressions.

### DIFF
--- a/sessions/js-conditions-and-booleans/truthy-and-falsy/js/index.js
+++ b/sessions/js-conditions-and-booleans/truthy-and-falsy/js/index.js
@@ -3,7 +3,6 @@
 
 
 function showAnswers() {
-    console.log("in ShowAnswers()");
     const spans = document.querySelectorAll("span");
     spans.forEach((span) => {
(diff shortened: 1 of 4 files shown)


sessions/js-conditions-and-booleans/truthy-and-falsy/js/index.js
```diff
@@ -1,2 +1,13 @@
-const truthyOrFalsy = Boolean(3 >= 2);
-console.log("The expression is " + truthyOrFalsy);
+// You don't have to touch this code.
+// You can solve this exercise in the browser
+
+
+function showAnswers() {
+    console.log("in ShowAnswers()");
```
MerleMak, line 6:
```suggestion
```

MerleMak (commented):
Nice Roland, i think this is a great improvement for the challenge, thank you!!

PR #364, opened 2024-03-13
Bad wording was leading students towards error.

alexxandar:
Students would enter only the cluster name, as per description and then would get an error "URI must include hostname, domain name, and tld". By calling it cluster URL we can avoid such missunderstanding.


ThomasKirst (approved):
Hi Alexxandar,

good catch! I´m not working for neuefische anymore, but in my opinion this is a reasonable improvement and I approve it :)

Cheers,
Thomas

F-Kirchhoff:
Hey Alex,
good point. I will merge the PR.

PR #373, opened 2024-04-26
fix: remove codesandbox links from challenge

F-Kirchhoff:


Roland-Hufnagel (changes requested):
Shouldn't we remove this part as well:

### CodeSandbox Cloud
Select the "Preview: 3000" tab to view this project.
Select the "Tests: logs" tab to view the tests.

And I would recommend to remove test-related content as well (since there are no tests).

F-Kirchhoff:
regarding the tests, it might be better to leave them in for now to remove them in bulk with a find / replace later on.

Roland-Hufnagel (approved):
LGTM

PR #385, opened 2024-07-08
split demo code into multiple files

F-Kirchhoff:
- split code into 3 files for better focus during the session
- show 3 different ways of writing callback functions in order for students to learn the concept callback function and not just the code pattern `forEach((item) => {...})`


JessicaLoers (approved):
**Hi @F-Kirchhoff** ✌️

I haven't held the session for a while, but I think it's a good improvement.  

A tiny note: it would be helpful if the change is also updated in the session guide. If you have to prepare on short notice it will be easier to find your feet. 

For example as a comment in die Codeblock

```js
⤵️
// numbers.js

const numberToCheck = 3333;
const numberIsIncluded = numbers.includes(numberToCheck);
numberOutput.textContent = numberIsIncluded
  ? `Yes, ${numberToCheck} is included!`
  : `No, ${numberToCheck} is not included`;
```

PR #396, opened 2024-09-02
Migrate/vite

F-Kirchhoff:


JessicaLoers (approved):
I went through all 969 files—just kidding! 🫠
I tested a few challenges at random, and they worked flawlessly. 

The following challenges were tested:
- react-basics-demo-start
- react-basics-attributes
- react-state-1-counter
- react-props-box-with-props.

From my side, everything’s approved!

PR #398, opened 2024-10-04
replace .js with .jsx endings in Readmes of vite projects

Roland-Hufnagel:
I discovered that the README.md files of the vite projects often referred to components or files with the '.js' Ending like 'You only have to touch the App.js'. This PR substitutes the '.js' with '.jsx' in the corresponding READMEs.


F-Kirchhoff (approved):
LGTM, thanks!

PR #399, opened 2024-10-04
Fix name on package js

mgczarnievicz:
Fix name in package json.
When doing for the first time `npm i` this error `npm ERR! code EDUPLICATEWORKSPACE must not have multiple workspaces with the same name` was happening due to the name was the same as demo start


Roland-Hufnagel (approved):
LGTM.

PR #403, opened 2024-10-08
Tourio App update bonus and CSS

Klausstille:
Update tourio app to include bonus material. Update example implementation to reflect bonus. Update to UI. Minor CSS update to make the app look fresher.


Klausstille:
Apologies for the numerous changes in this PR. It'd be best to pull the branch locally and review both the Tourio app template and the example solution.

PR #404, opened 2024-10-08
Add/js unit testing solutions

NikRoe:


sessions/js-unit-testing/math-functions_solution/index.test.js
```diff
@@ -0,0 +1,67 @@
...
+    const result = multiply(2, -4);
+
+    expect(result).toBeLessThan(0);
+});
+
+test("returns a positive value if called with two negative arguments", () => {
+    const result = multiply(-2, -4);
+
+    expect(result).toBeGreaterThan(0);
+});
```
F-Kirchhoff, line 55: I was thinking that we should show an example where multiple expect statements are in one test. Since these all have the same structure students might come to the conclusion that you have to obey this "one test one expect" structure.

F-Kirchhoff (approved):
Hey Niklas,
LGTM! I only had one general question how the structure the challenge. What do you think about it?

sessions/js-unit-testing/math-functions_solution/index.test.js
NikRoe: Hi Felix,

I think it is a good point, but it would mean adjusting the challenge.

Currently we give them the criteria to test for and we ask them to write a test for each of them.

In general, I like having separate well documented tests for complex tests and with simple tests as in this challenge, it makes sense to combine them into one test. We could even consider adjusting the demo a little bit, to have one test with several expects.

I would propose this: 
I would merge the current solutions, since they match what we currently show students.
I would also create a small ticket to adjust the Unit Testing Session, so that we show a test with several expects in the demo and also with an adjustment to the math challenge. 
What do you think?

F-Kirchhoff: that sounds perfect, lets do it like that.

PR #405, opened 2024-10-08
Tourio App update bonus and CSS

Klausstille:
Update tourio app to include bonus material. Update example implementation to reflect bonus. Update to UI. Minor CSS update to make the app look fresher.


sessions/recap-project-6/tourio-app/pages/api/places/[id]/index.js
```diff
@@ -7,11 +8,16 @@ export default function handler(request, response) {
     return;
   }
 
-  const place = places.find((place) => place.id === id);
+  const place = db_places.find((place) => place._id.$oid === id);
+  const comment = place?.comments;
+  const allCommentIds = comment?.map((comment) => comment.$oid) || [];
+  const comments = db_comments.filter((comment) =>
+    allCommentIds.includes(comment._id.$oid)
+  );
```
Comment (F-Kirchhoff) [inline sessions/recap-project-6/tourio-app/pages/api/places/[id]/index.js:16]: So you implemented the comments in the "non db" version as well. In my opinion this can confuse the students a lot. From my side the starter template should not include any elements regarding the comments (except the Comments Component maybe). By this we don't need the comments included in the data from the beginning, preventing the need to the awkward Object Ids. 
I would suggest that we keep your example solution and the starter template from the old project. We could think about providing example data with the example solution / bonus part to get the students going. They could replace their current data in their DB with the new data with comments as soon as they reach the bonus challenge. In this way, the students can focus on the first tasks and can shift towards the bonus challenge if they want to later.

F-Kirchhoff (changes requested):
Hey Klaus,
unfortunately the spiced version of this challenge is a bit out of date (see styled components). There are also a bunch of smaller issues. 
I would like to propose a different approach to the bonus challenge with the comments. (see my last comment).

sessions/recap-project-6/tourio-app/README.md
```diff
@@ -26,9 +26,23 @@ In this project you will build a Tourio App, the best friend of all travelers: i
...
+
+- Write a POST API route to handle comment submissions. This route should associate each comment with the appropriate place in your database.
+
+- Implement a mechanism to retrieve, display and delete comments for each place on the detail page.
+
+- Consider the case where a place is deleted. Ensure associated comments are also removed to avoid orphaned data in your database.
+
+### Deploy your Project to Vercel
+
+Follow [these instructions](https://github.com/neuefische/web-protocol-template/blob/main/docs/deployment-vercel.md) to deploy the project to Vercel.
```
Klausstille, line 45: @F-Kirchhoff this would mean our students would need to have access to neuefische repo too.

F-Kirchhoff: you are absolutely correct. We probably need to move the curriculum and exercises into a neutral organisation, so we could move these docs there as well as soon we decided on how we want to share our common resources with the cohorts.

sessions/recap-project-6/example-implementation-with-comments/components/StyledLink.js
```diff
@@ -0,0 +1,25 @@
+import styled, { css } from "styled-components";
+
+export const StyledLink = styled.a`
```
F-Kirchhoff, line 3: we can use Link here instead
```suggestion
import Link from 'next/link'

export const StyledLink = styled(Link)`
```

sessions/recap-project-6/example-implementation-with-comments/pages/index.js
```diff
@@ -0,0 +1,42 @@
...
+              image={place.image}
+              location={place.location}
+              id={place._id}
+            />
+          );
+        })}
+      </ListContainer>
+      <Link href="/create" passHref legacyBehavior>
+        <FixedLink>+ place</FixedLink>
+      </Link>
```
F-Kirchhoff, line 39: Here we could change the nested links

sessions/recap-project-6/example-implementation-with-comments/pages/places/[id]/edit.js
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  if (!isReady || isLoading || error) return <h2>Loading...</h2>;
+
+  return (
+    <>
+      <h2 id="edit-place">Edit Place</h2>
+      <Link href={`/places/${id}`} passHref legacyBehavior>
+        <StyledLink justifySelf="start">back</StyledLink>
+      </Link>
```
Comment (F-Kirchhoff) [inline sessions/recap-project-6/example-implementation-with-comments/pages/places/[id]/edit.js:36]: here as well

sessions/recap-project-6/example-implementation-with-comments/pages/api/comments/[id]/index.js
```diff
@@ -0,0 +1,37 @@
+import Comment from "../../../../db/models/Comment";
+import dbConnect from "../../../../db/connect";
+
+export default async function handler(request, response) {
+  const { id } = request.query;
+  await dbConnect();
+
+  if (request.method === "GET") {
+    try {
+      const comments = await Comment.find({ place: id });
```
Comment (F-Kirchhoff) [inline sessions/recap-project-6/example-implementation-with-comments/pages/api/comments/[id]/index.js:10]: here we can use placeId as well, it would be a bit more explicit

sessions/recap-project-6/example-implementation-with-comments/db/models/Comment.js
```diff
@@ -0,0 +1,11 @@
+import mongoose from "mongoose";
+const { Schema } = mongoose;
+const commentSchema = new Schema({
+  name: { type: String, required: true },
+  comment: { type: String, required: true },
+  place: { type: Schema.Types.ObjectId, ref: "Place" },
```
F-Kirchhoff, line 6:
```suggestion
  placeId: { type: Schema.Types.ObjectId, ref: "Place" },
```

sessions/recap-project-6/example-implementation-with-comments/components/Comments.js
```diff
@@ -0,0 +1,101 @@
...
+
+  async function handleDeleteComment(comment_id) {
+    const response = await fetch(
+      `/api/comments/${id}?comment_id=${comment_id}`,
+      {
+        method: "DELETE",
+      }
+    );
+    if (response.ok) {
+      await response.json();
```
F-Kirchhoff, line 63: this is not needed

sessions/recap-project-6/example-implementation-with-comments/components/Comments.js
```diff
@@ -0,0 +1,101 @@
...
+        "Content-Type": "application/json",
+      },
+      body: JSON.stringify({ ...commentData, place: id }),
+    });
+    if (response.ok) {
+      await response.json();
+      mutate();
+      e.target.reset();
+    } else {
+      console.error(`Error: ${response.status}`);
```
F-Kirchhoff, line 51:
```suggestion
  async function handleSubmitComment(event) {
    event.preventDefault();
    const formData = new FormData(event.target);
    const commentData = Object.fromEntries(formData);
    const response = await fetch(`/api/comments/${id}`, {
      method: "POST",
      headers: {
        "Content-Type": "application/json",
      },
      body: JSON.stringify({ ...commentData, place: id }),
    });
    if (response.ok) {
      await response.json();
      mutate();
      event.target.reset();
    } else {
      console.error(`Error: ${response.status}`);
```

sessions/recap-project-6/example-implementation-with-comments/components/Comments.js
```diff
@@ -0,0 +1,101 @@
...
+  flex-direction: column;
+  align-items: center;
+  border: 1px solid black;
+  border-radius: 0.8rem;
+  padding: 0.5rem;
+  text-align: center;
+  p {
+    border-bottom: solid 1px black;
+    padding: 20px;
+  }
```
F-Kirchhoff, line 19: We teach our students to avoid nested css rulesets, since it changes the specificity of the rule. I think this should be its own styled component.

sessions/recap-project-6/example-implementation-with-comments/components/Card.js
```diff
@@ -0,0 +1,47 @@
...
+  position: relative;
+  margin: 0;
+  padding: 0;
+  line-height: 0.2rem;
+`;
+
+export default function Card({ name, image, location, id }) {
+  return (
+    <Article>
+      <Link href={`places/${id}`} passHref legacyBehavior>
```
F-Kirchhoff, line 27: passHref and legacyBehavior should not work in this scenario.

sessions/recap-project-6/example-implementation-with-comments/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+4. Redeploy your application:
+
+   1. In the main navigation, choose "Deployments".
+   2. Open the three dots next to your last deployment and choose "Redeploy".
+
+5. If there's a popup, hit the "Redeploy" button again.
+
+6. Congratulations, you are done! Open the Vercel URL of your project to see that your deployed application has now access to the cloud database.
+
```
F-Kirchhoff, line 75: Here we should use the same readme as in our new starter template.

F-Kirchhoff (changes requested):
Hey Klaus,
great work, thanks a lot. There is one fundamental issue I would like to discuss, whether we implement the comments in the starter template or not. I would suggest not to do it and only keep the Comments Component without using it in the starter. Otherwise there might be issues if we migrate the places to the mongoDB and the ids change, then the comments will not be associated with the places anymore. Further, the comments should be a bonus challenge, if its baked into the starter files it does not really feel like a bonus. What do you think?

F-Kirchhoff (approved):

Klausstille (commented):
Great! 👌

PR #407, opened 2024-10-14
js-conditions-and-booleans: new challenge

JessicaLoers:
- Moved "old" unused challenge "booleans" to archive folder.
- Added a new challenge: "Weather check" using a `switch` statement.
- Added solution for the new challenge.
- Replaced the previous "LogType" challenge with this new one.

**Reason for Change:** 
The new "Weather check" challenge provides a simpler way for students to learn and practice the syntax of the switch statement. Unlike the previous "LogType" challenge, which required understanding more complex concepts like `typeof` and different data types, this challenge focuses on just basic conditions.  
**_This allows students to focus on mastering the core `switch` statement without added complexity._**


sessions/js-conditions-and-booleans/weather-check_solution/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "js-conditions-and-booleans_weather-check",
```
mgczarnievicz, line 2:
```suggestion
  "name": "js-conditions-and-booleans_weather-check-solution",
```

mgczarnievicz (commented):
I would make them run the challenge `Weather check` with node in the terminal.
They don't need the html files and they need practice with node.

mgczarnievicz:
In this challenge, we intentionally had the students use the `typeof` operator to help them become familiar with checking the type of data in JavaScript. They gain an essential skill for debugging and understanding the data they are working with. The goal was that next time they encounter an unfamiliar data type, they’ll know how to quickly determine its type using `typeof`, making their coding process smoother.

Yes, this challenge combined multiple areas of knowledge, making it more difficult. While it may have pushed students a bit further, it also served as a recap of the different data types in JavaScript. This way, they not only strengthened their knowledge of `typeof` but also revisited core concepts, enhancing their problem-solving skills in more complex scenarios.

I am okay with taking it out, if you feel is too much for the students.

JessicaLoers:
Hey @mgczarnievicz, thank you for the review.

- It's a good point about having them run the Weather check challenge with Node in the terminal. 

- I understand your point and I also think the "logType" challenge is valuable. However, I believe that for the initial understanding of the `switch` statement, it would be more helpful to focus on the essentials, especially since we aren’t covering the switch statement in the session.

My suggestion would be to offer the "logtype" challenge as a bonus with a higher difficulty level.

What do you think?

mgczarnievicz:
> My suggestion would be to offer the "logtype" challenge as a bonus with a higher difficulty level.

Yes let, do this.
We can ask them to run this one two in the terminal with node.

PR #410, opened 2024-10-18
Merge/backend route handling

F-Kirchhoff:
- use consistent route handling syntax
- add error handling to backend update and delete demo end 
- simplify codebase by reducing components and folder nesting.

For testing a working mongoDB instance is needed with the following databases:
- jokes-database with jokes collection
- fish-shop with products and comments 

Don't forget to use an .env.local or the following command before starting the dev server:
```bash
export MONGODB_URI="your connection string"
```


NikRoe (changes requested):
Hi Felix,

I like the changes a lot. It makes the codebase much easier to read and comprehend, the consistency is also a big plus.

Testing the adjusted demos and challenges also worked nicely 👍 

I only noticed two `export` statement that I believe are not needed. Then this is PR is good to go.

However, I would wait before merging this, since the session guide no longer matches the demos perfectly.

I will create a PR for the web-curriculum to update the session guides to reflect the new demos.

NikRoe (approved):
👍

PR #413, opened 2024-10-25
Merge/js modern syntax

JessicaLoers:
- Added "Back to Basics" challenge (formerly "retro challenge") with updated instructions, test, and solution.
- Integrated the 3rd challenge from the Spiced "Destructuring Objects" series as Exercise 5, along with corresponding test and solution.


mgczarnievicz (approved):
👏 👌🏼

PR #420, opened 2024-11-08
Adjust challenge to better fit the session content: JS DOM and Events: - Simple Calculator

ahohnsen:
I noticed that the **Simple Calculator** challenge from the **JS DOM and Events** session doesn’t really match the session content - it actually fits better with **JS Basics and Variables**. The differences may have come up due to the merge process. 
Since the **JS Basics and Variables** session already has enough challenges, I’ve adjusted the **Simple Calculator** challenge so that students can review the main topics of the session, especially `document.querySelector` and `addEventListener`.
Let me know, what you think. 🙂


sessions/js-dom-and-events/simple-calculator/README.md
```diff
@@ -1,7 +1,18 @@
-# simple-calculator
+# JS DOM and Events: Simple Calculator
+
+In this challenge, you’ll use JavaScript to add interactivity to a simple calculator. You’ll work with `document.querySelector` and `addEventListener` to perform calculations when buttons are clicked.
 
 ## Task
```
F-Kirchhoff, line 5:
```suggestion
## Tasks
```

sessions/js-dom-and-events/simple-calculator/README.md
```diff
@@ -1,7 +1,18 @@
...
 
+1. **Examine the HTML**: Open the `index.html` file and take a look at the structure of the calculator. Notice that each button corresponds to a specific operation or action, but the calculator currently has no functionality.
+
+2. **Add Functionality**: Open the `./js/index.js` file and complete each step as described in the comments.
+
+   - **Mathematical Operations**: Write code to perform each mathematical operation (addition, subtraction, etc.) when its corresponding button is clicked. Each operation should use `operand1` and `operand2`, then log the result to the console.
+   - **Update the First Operand**: Implement functionality to update the value of `operand1` using the buttons in the "Update the First Operand" section. Each button should modify `operand1` and log the updated value.
+
+Switch to the `./js/index.js` file to complete each task as described.
+
```
F-Kirchhoff, line 15: I like the text, with the enumerated bullet points with the bold text it looks a bit much like it was generated by chatgpt. Maybe we can adjust it to the other task descriptions? What do you think?

F-Kirchhoff (approved):
I think these are great additions, or better said subtractions. By this they practice the dom methods more which is what they should do. I only have a question regarding the style of the tasks description.

sessions/js-dom-and-events/simple-calculator/README.md
ahohnsen: I actually noticed just now, that this is not consistent in our challenges - sometimes we use "Tasks" or "Task". 🙁 I can change it everywhere to "Tasks" as it would make more sense. Right?

sessions/js-dom-and-events/simple-calculator/README.md
ahohnsen: Yes, I had some help from ChatGPT. Anyhow, I decided to write it like this as this is more of a summary of the main exercises. 🤔 The detailed instructions are in the `index.js`. This is also not consistent throughout the challenges. 🫣

ahohnsen: I updated both.

sessions/js-dom-and-events/simple-calculator/README.md
ahohnsen: I replaced "Task" everywhere with "Tasks".

